PR #2, opened 2020-11-10
feat: support string array in OAuth aud claim

emilhdiaz:
Description of changes:

As per [rfc7519](https://tools.ietf.org/html/rfc7519#section-4.1.3): 

> In the general case, the "aud" value is an array of case-
   sensitive strings, each containing a StringOrURI value.  In the
   special case when the JWT has one audience, the "aud" value MAY be a
   single case-sensitive string containing a StringOrURI value.

This change allows the `aud` claim in the `access_token` JWT to be either a `string` or `string[]` (string array). 

### DIFF
--- a/src/smartConfig.ts
+++ b/src/smartConfig.ts
@@ -21,7 +21,7 @@ export interface SMARTConfig {
     /**
      * Per SMART spec this is the 'aud' key found in the access_token
      */
-    expectedAudValue: string; //
+    expectedAudValue: string;
     /**
      * Per SMART spec this is the 'iss' key found in the access_token
      */
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -39,11 +39,14 @@ export class SMARTHandler implements Authorization {
         // Decoding first to determine if it passes scope & claims check first
         const decoded = decode(request.accessToken, { json: true }) || {};
         const { aud, iss } = decoded;
+        const audArray = Array.isArray(aud) ? aud : [aud];
+
         // verify aud & iss
-        if (this.config.expectedAudValue !== aud || this.config.expectedIssValue !== iss) {
+        if (!audArray.includes(this.config.expectedAudValue) || this.config.expectedIssValue !== iss) {
             console.error('aud or iss is not matching');
             throw new UnauthorizedError('Error validating the validity of the access_token');
         }
+
         // verify scope
         let scopes: string[] = [];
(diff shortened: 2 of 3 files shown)


src/smartConfig.ts
```diff
@@ -21,7 +21,7 @@ export interface SMARTConfig {
     /**
      * Per SMART spec this is the 'aud' key found in the access_token
      */
-    expectedAudValue: string; //
+    expectedAudValue: string | string[];
```
rsmayda: This value signifies what we should look for in the `aud` claim. We know that the `aud` claim _must_ contain this fhir server url

So I think it makes sense for this to be a single string

src/smartHandler.ts
```diff
@@ -39,11 +40,13 @@ export class SMARTHandler implements Authorization {
         // Decoding first to determine if it passes scope & claims check first
         const decoded = decode(request.accessToken, { json: true }) || {};
         const { aud, iss } = decoded;
+
         // verify aud & iss
-        if (this.config.expectedAudValue !== aud || this.config.expectedIssValue !== iss) {
+        if (!isEqual(this.config.expectedIssValue, iss) || !isEqual(this.config.expectedAudValue, aud)) {
```
rsmayda: `aud` here could be an array since the access_token could be giving access to many EHRs

src/smartConfig.ts
emilhdiaz: @rsmayda good point. So would it make more sense to keep `expectedAudValue` here as a string, but perform a string match OR array lookup in the handler?

rsmayda: Yep that is what I am thinking, we should check it against the access_token's `aud` being a string or an array!

emilhdiaz: Makes sense, I reverted the SMARTConfig changes and instead updated the `aud` check to assume either string or array.

src/smartHandler.ts
emilhdiaz: I reverted the `isEqual()` calls here given I no longer need to compare the equality of array.

rsmayda (approved):
Looks good!

rsmayda (changes requested):
Whoops forgot to ask can you add a test case for when aud is an array?

emilhdiaz:
Ah yep, just did. It's my first time writing unit tests in jest, so hopefully I implemented the tests correctly.

rsmayda (approved):
Unit tests look good! Thanks

PR #3, opened 2020-11-10
Auth strategy

emilhdiaz:
**Background:** 

The SMART-on-FHIR specification only governs the exchange of scopes between the SMARTApp and the authorization server and does not take a position on how an authorization server and resource server ought to coordinate authorization. The latter is left up to the implementer. 

The current implementation uses token introspection via the `/userinfo` url. However, The SMART-on-FHIR specification does not require the `fhirUser` claim to be present in the `access_token` or via the `/userinfo` endpoint on the authorization server. Therefore, other implementations of the authorization server (like Auth0) will not be able to comply with these expectations. 

This is a proposal for a more flexible strategy pattern that will allow implementers to design their own token validation and authorization strategies.

**Description of changes:**

This change refactors the `SMARTHandler.isAuthorized()` method such that the specific authorization strategy used to implement SMART-on-FHIR can be encapsulated in classes that implement the (new) `AuthStrategy` interface. This allows other implementations to define their own strategies and inject them into the authorization flow via the `SMARTConfig` object. 

Example config: 
```
const authZConfig: SMARTConfig = {
    version: 1.0,
    scopeKey: 'scp',
    scopeValueType: 'array',
    scopeRule,
    expectedAudValue: expectedAud,
    expectedIssValue: expectedIss,
    expectedFhirUserClaimKey: 'fhirUser',
    fhirUserClaimRegex: /(\w+)\/(\w+)/g,
    authZUserInfoUrl: `${expectedIss}/userInfo`,
    authStrategies: [new UserInfoStrategy(`${expectedIss}/userInfo`, 'fhirUser')],
};
```

### DIFF
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -64,7 +64,7 @@ export class SMARTHandler implements Authorization {
         const results = [];
         for (let i = 0; i < this.config.authStrategies.length; i += 1) {
             const strategy = this.config.authStrategies[i];
-            results.push(strategy.isTokenValid(request.accessToken));
+            results.push(strategy.validateToken(request.accessToken));
         }
         await Promise.all(results);
     }
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -14,7 +14,6 @@ import {
     ReadResponseAuthorizedRequest,
     WriteRequestAuthorizedRequest,
 } from 'fhir-works-on-aws-interface';
-import axios from 'axios';
 import { LaunchType, ScopeType, SMARTConfig } from './smartConfig';
 
 // eslint-disable-next-line import/prefer-default-export
@@ -62,20 +61,12 @@ export class SMARTHandler implements Authorization {
         }
 
         // Verify token
-        let response;
-        try {
-            response = await axios.post(
-                this.config.authZUserInfoUrl,
-                {},
-                { headers: { Authorization: `Bearer ${request.accessToken}` } },
-            );
-        } catch (e) {
-            console.error('Post to authZUserInfoUrl failed', e);
+        const results = [];
+        for (let i = 0; i < this.config.authStrategies.length; i += 1) {
+            const strategy = this.config.authStrategies[i];
+            results.push(strategy.isTokenValid(request.accessToken));
         }
-        if (!response || !response.data[this.config.expectedFhirUserClaimKey]) {
-            console.error(`result from AuthZ did not have ${this.config.expectedFhirUserClaimKey} claim`);
-            throw new UnauthorizedError("Cannot determine requester's identity");
-        }
+        await Promise.all(results);
     }
 
     async isBundleRequestAuthorized(request: AuthorizationBundleRequest): Promise<void> {
@@ -100,7 +91,7 @@ export class SMARTHandler implements Authorization {
         await this.isAuthorized({ accessToken: request.accessToken, operation: request.operation });
     }
 
-    private isScopeSufficient(
+    protected isScopeSufficient(
         scopes: string[],
         operation: TypeOperation | SystemOperation,
         resourceType?: string,
@@ -140,7 +131,7 @@ export class SMARTHandler implements Authorization {
         return false;
     }
 
-    private getValidOperation(scopeType: ScopeType, accessType: string) {
+    protected getValidOperation(scopeType: ScopeType, accessType: string) {
         let validOperations: (TypeOperation | SystemOperation)[] = [];
         if (accessType === '*' || accessType === 'read') {
             validOperations = this.config.scopeRule[scopeType].read;
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -14,6 +14,7 @@ import {
     ReadResponseAuthorizedRequest,
     WriteRequestAuthorizedRequest,
 } from 'fhir-works-on-aws-interface';
+import axios from 'axios';
 import { LaunchType, ScopeType, SMARTConfig } from './smartConfig';
 
 // eslint-disable-next-line import/prefer-default-export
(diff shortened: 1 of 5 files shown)


src/smartStrategy.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface AuthStrategy {
+    /**
+     * Validates the JWT access_token
+     * @param accessToken
+     * @throws UnauthorizedError
+     */
+    isTokenValid(accessToken: string): Promise<void>;
+}
+
+export class TokenIntrospectionStrategy implements AuthStrategy {
```
rsmayda:
```suggestion
export class UserInfoStrategy implements AuthStrategy {
```

src/smartStrategy.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface AuthStrategy {
+    /**
+     * Validates the JWT access_token
+     * @param accessToken
+     * @throws UnauthorizedError
+     */
+    isTokenValid(accessToken: string): Promise<void>;
+}
+
```
rsmayda, line 14: Add a comment explaining this strategy; something like:
```suggestion
/**
  * This strategy uses the /userinfo endpoint to verify the `access_token` and get the `fhirUser` claim for further authorization. userinfo endpoint is OIDC compliant: https://openid.net/specs/openid-connect-core-1_0.html#UserInfo
  */
```

src/smartStrategy.ts
```diff
@@ -0,0 +1,39 @@
+import axios from 'axios';
+import { UnauthorizedError } from 'fhir-works-on-aws-interface';
+
+export interface AuthStrategy {
+    /**
+     * Validates the JWT access_token
+     * @param accessToken
+     * @throws UnauthorizedError
+     */
+    isTokenValid(accessToken: string): Promise<void>;
```
rsmayda: So we are using userinfo to get user identity information perhaps a better method name:
```suggestion
     * @throws UnauthorizedError
     * @returns `fhirUser` claim
     */
    validateToken(accessToken: string): Promise<string>;
```

src/smartHandler.ts
```diff
@@ -97,7 +91,7 @@ export class SMARTHandler implements Authorization {
         await this.isAuthorized({ accessToken: request.accessToken, operation: request.operation });
     }
 
-    private isScopeSufficient(
+    protected isScopeSufficient(
```
rsmayda: any reason to make these protected vs private?

src/smartHandler.ts
```diff
@@ -59,20 +61,12 @@ export class SMARTHandler implements Authorization {
...
-        }
-        if (!response || !response.data[this.config.expectedFhirUserClaimKey]) {
-            console.error(`result from AuthZ did not have ${this.config.expectedFhirUserClaimKey} claim`);
-            throw new UnauthorizedError("Cannot determine requester's identity");
+        const results = [];
+        for (let i = 0; i < this.config.authStrategies.length; i += 1) {
+            const strategy = this.config.authStrategies[i];
+            results.push(strategy.isTokenValid(request.accessToken));
         }
+        await Promise.all(results);
```
rsmayda: You should be able to streamline this like: 
```suggestion
        const results = this.config.authStrategies.map( strategy => { return strategy(accessToken) } );
        await Promise.all(results);
```

rsmayda (commented):
Great start and bringing of new ideas. Can you provide some use-cases to when a user will want to use multiple strategies?

src/smartHandler.ts
emilhdiaz: This was left over from a different approach I was experimenting with.

src/smartStrategy.ts
emilhdiaz: Good suggestion, renamed

src/smartStrategy.ts
emilhdiaz: Thanks for the suggestion, added!

src/smartStrategy.ts
emilhdiaz: Good suggestion, renamed.

src/smartHandler.ts
emilhdiaz: Much cleaner! Thanks.

emilhdiaz:
The use cases I had in mind for multiple strategies:

- We can use different strategies for different types of users of the API. For example, one strategy can be implemented for a patient accessing their own data, another for an authorized representative accessing data on behalf of a patient, and a third for a hospital or plan admin accessing data on their patient population. Each strategy could skip validation altogether if the access context doesn't match, but the authorization module would be aware of all the available strategies for each access context. 

-  This would allow us to implement strategies in a modular fashion and then combine them to form more complex strategies. For example, one strategy could validated clinical scopes against the operations permitted for those scopes, while another strategy can validate fine-grain access control against the specific resource being requested, while yet a final strategy can layer in vertical access control against the specific attributes being requested within the resource. 

Now, I admit these are a bit contrived, but I also figured adding the flexibility now couldn't hurt. Thoughts?

src/smartHandler.ts
emilhdiaz: Reverted this change.

src/smartStrategy.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    async validateToken(accessToken: string): Promise<void> {
+        let response;
+        try {
+            response = await axios.post(
+                this.authZUserInfoUrl,
+                {},
+                { headers: { Authorization: `Bearer ${accessToken}` } },
+            );
```
rsmayda: So the benefit of using the /userinfo endpoint is that you get the `fhirUser` claim from the endpoint. that would in the `response` above.

Perhaps the flow needs to be:
1. aud/iss check
2. scope check
3. validateToken and get `fhirUser` claim
4. auth strategies check

Thoughts?

emilhdiaz: Originally, I intended to pass the `fhirUser` claim through the `access_token` (in addition to the `id_token` as required by the SoF specs), but the more I think about it, pulling it from the `/userinfo` endpoint makes more sense. In our case, it will require us to open up our outbound firewall to allow the resource server to communicate with the authorization server.

rsmayda: Sounds good!

Just to lay out all options there are three ways that I know of to verify the access_token + get `fhirUser` claim
1. Put `fhirUser` inside the `access_token`
   1. Verify `access_token` (locally) by using the your Oauth2's public key
   1. Verify `access_token` (remotely) by using your Oauth2's [token introspection endpoint](https://www.oauth.com/oauth2-servers/token-introspection-endpoint/)
1. Use [user info endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) and get the `fhirUser` claim from Oauth2 server

rsmayda:
I am going to archive this thought for now (as we have tabled it and this PR is now stale) -- but happy to revisit if interested

PR #4, opened 2020-11-18
feat: implement new interface

rsmayda:
### `verifyAccessToken` changes
- Expected `fhirUser` regex is now a constant `FHIR_USER_REGEX`
  - `fhirUser` is expected to be an absolute url
  - If `fhirUser` claim is not in the right format we reject
- metadata & .well_known changed to pass authorization
- return the userIdentity found by the call to userInfo

### `authorizeAndFilterReadResponse` changes
- For searching filter out resources user does not have access to, does NOT throw error
  - Thought here is if all elements are filtered out the router should call for the next page of search results and try again
- For single reads, throw error if user is not authorized
- For external users (ie their resource is not from our FHIR server) they must use the fully qualified reference to have access
- To check if user is a reference I am converting the JSON to a string then checking for the `fhirUser` claim
- Internal Practitioners can access anything

### `isWriteRequestAuthorized` changes
- Make this Practitioner only action

### DIFF
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -18,7 +18,7 @@ import {
     KeyValueMap,
 } from 'fhir-works-on-aws-interface';
 import axios from 'axios';
-import { LaunchType, ScopeType, SMARTConfig } from './smartConfig';
+import { IdentityType, LaunchType, ScopeType, SMARTConfig } from './smartConfig';
 
 // eslint-disable-next-line import/prefer-default-export
 export class SMARTHandler implements Authorization {
@@ -35,6 +35,8 @@ export class SMARTHandler implements Authorization {
         'search-system',
         'history-system',
     ];
+
+    private readonly typesWithWriteAccess: IdentityType[] = ['Practitioner'];
 
     private readonly version: number = 1.0;
 
@@ -143,12 +145,15 @@ export class SMARTHandler implements Authorization {
 
     // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
     async isWriteRequestAuthorized(request: WriteRequestAuthorizedRequest): Promise<void> {
-        // TODO this is stubbed for now
+        const fhirUser = this.getFhirUser(request.userIdentity);
+        if (fhirUser.hostname !== this.apiUrl || !this.typesWithWriteAccess.includes(fhirUser.resourceType)) {
+            throw new UnauthorizedError('User does not have permission for requested operation');
+        }
     }
 
     // eslint-disable-next-line class-methods-use-this
     private authorizeResource(
-        fhirUser: { hostname: string; resourceType: string; id: string },
+        fhirUser: { hostname: string; resourceType: IdentityType; id: string },
         resource: any,
     ): boolean {
         const jsonStr = JSON.stringify(resource);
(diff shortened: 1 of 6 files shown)


src/smartHandler.ts
```diff
@@ -5,36 +5,61 @@
...
-import { LaunchType, ScopeType, SMARTConfig } from './smartConfig';
+import { IdentityType, LaunchType, ScopeType, SMARTConfig } from './smartConfig';
 
 // eslint-disable-next-line import/prefer-default-export
 export class SMARTHandler implements Authorization {
     static readonly CLINICAL_SCOPE_REGEX = /^(?<scopeType>patient|user|system)\/(?<scopeResourceType>[a-zA-Z]+|\*)\.(?<accessType>read|write|\*)$/;
 
     static readonly LAUNCH_SCOPE_REGEX = /^(?<scopeType>launch)(\/(?<launchType>patient|encounter))?$/;
 
+    static readonly FHIR_USER_REGEX = /(?<hostname>(http|https):\/\/([A-Za-z0-9\-\\.:%$]*\/)+)(?<resourceType>Person|Practitioner|RelatedPerson|Patient)\/(?<id>[A-Za-z0-9\-.]+)$/;
```
nguyen102: Can we add some tests for this REGEX? It'll help us document what we expect this REGEX to match. It'll also confirm that the REGEX works as expected.

src/smartHandler.ts
```diff
@@ -54,53 +79,118 @@ export class SMARTHandler implements Authorization {
...
+
+    private getFhirUser(userIdentity: KeyValueMap): { hostname: string; resourceType: IdentityType; id: string } {
+        const { fhirUserClaimKey } = this.config;
+        const fhirUserValue = userIdentity[fhirUserClaimKey];
+        const match = fhirUserValue.match(SMARTHandler.FHIR_USER_REGEX);
+        if (match) {
+            const { hostname, resourceType, id } = match.groups!;
+            return { hostname, resourceType, id };
+        }
+        throw new UnauthorizedError("Requester's identity is in the incorrect format");
```
nguyen102, line 192: Do we want to pass along the `userIdentity` in the error message for context?

src/smartHandler.ts
```diff
@@ -54,53 +79,118 @@ export class SMARTHandler implements Authorization {
...
+        return R4_PATIENT_COMPARTMENT_RESOURCES;
     }
 
     async authorizeAndFilterReadResponse(request: ReadResponseAuthorizedRequest): Promise<any> {
-        // TODO this is stubbed for now
-        await this.isAuthorized({ accessToken: request.accessToken, operation: request.operation });
-        return request.readResponse;
+        const fhirUser = this.getFhirUser(request.userIdentity);
+        const { operation, readResponse } = request;
+        if (this.searchOperations.includes(operation)) {
```
nguyen102, line 132: What is the list of `searchOperations` for, are those operations that we support for SMART? Should that a be a config value, or is it always a set list?

Also the list does not include the `read` operation. Does that mean a requested won't be able to read any resources?

[Reference in Router](https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/routes/genericResourceRoute.ts#L40)

nguyen102 (dismissed):
Looks good. Some minor comments.

src/smartHandler.ts
rsmayda: +1 good call out and practice; found some minor additions

src/smartHandler.ts
rsmayda: I am hesitant as it may contain PII -- I think better to log/return less

nguyen102 (approved):

PR #5, opened 2020-11-19
feat: No auth needed for  route and  route

nguyen102:
Issue #, if available:

Description of changes:

`metadata` and `well-known` route doesn't need to be Authenticated/Authorized


src/smartHandler.ts
```diff
@@ -37,6 +37,13 @@ export class SMARTHandler implements Authorization {
     }
 
     async verifyAccessToken(request: VerifyAccessTokenRequest) {
+        if (
```
rsmayda, line 41: Add test case for this

nguyen102: updated

rsmayda (approved):

PR #7, opened 2020-12-03
[feat] Add Auth support for Bundles, Bulk Data, and allowedResourcesForAuth

nguyen102:
Issue #, if available:

Description of changes:
An overview of the major design choices in this PR is listed below.

* Adding scopes to the `userIdentity` object returned by `verifyAccessToken`. `scopes` is needed later on by `isBundleRequestAuthorized` and `getAllowedResourceTypesForOperation`. These two methods are called in the `handler` classes of the `router` package. 

* Added `operationToAllowedResources` to the constructor because `getAllowedResourceTypesForOperation` needs to know what operations are allowed by each resource. So for example the operation `search-type` can be performed on resources: Patient, Observation, etc...

* Functions implemented `isBundleRequestAuthorized`, `getAllowedResourceTypesForOperation`, and `isAccessBulkDataJobAllowed` 

### DIFF
--- a/src/smartHandler.test.ts
+++ b/src/smartHandler.test.ts
@@ -78,6 +78,10 @@ const launchEncounterAccess: string =
     'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2ZXIiOjEsImp0aSI6IkFULjZhN2tuY1RDcHUxWDllbzJRaEgxel9XTFVLNFR5VjQzbl85STZrWk53UFkiLCJpc3MiOiJodHRwczovL2Rldi02NDYwNjExLm9rdGEuY29tL29hdXRoMi9kZWZhdWx0IiwiYXVkIjoiYXBpOi8vZGVmYXVsdCIsImlhdCI6MTYwMzExODEzOCwiZXhwIjoxNjAzMTIxNzM4LCJjaWQiOiIwb2E4bXVhektTeWs5Z1A1eTVkNSIsInVpZCI6IjAwdTg1b3p3ampXUmQxN1BCNWQ1Iiwic2NwIjpbImZoaXJVc2VyIiwib3BlbmlkIiwicHJvZmlsZSIsImxhdW5jaC9lbmNvdW50ZXIiXSwic3ViIjoic21heWRhNDRAZ21haWwuY29tIiwiZmhpclVzZXIiOiJQYXRpZW50LzEyMzQifQ.Nl6yRFGCFnSXszg9kBHS4sbMQio7YnOUajGWOkaLdUA';
 const manyReadAccess: string =
     'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2ZXIiOjEsImp0aSI6IkFULjZhN2tuY1RDcHUxWDllbzJRaEgxel9XTFVLNFR5VjQzbl85STZrWk53UFkiLCJpc3MiOiJodHRwczovL2Rldi02NDYwNjExLm9rdGEuY29tL29hdXRoMi9kZWZhdWx0IiwiYXVkIjoiYXBpOi8vZGVmYXVsdCIsImlhdCI6MTYwMzExODEzOCwiZXhwIjoxNjAzMTIxNzM4LCJjaWQiOiIwb2E4bXVhektTeWs5Z1A1eTVkNSIsInVpZCI6IjAwdTg1b3p3ampXUmQxN1BCNWQ1Iiwic2NwIjpbImZoaXJVc2VyIiwib3BlbmlkIiwicHJvZmlsZSIsImxhdW5jaC9lbmNvdW50ZXIiLCJwYXRpZW50L1BhdGllbnQucmVhZCIsInBhdGllbnQvT2JzZXJ2YXRpb24ucmVhZCJdLCJzdWIiOiJzbWF5ZGE0NEBnbWFpbC5jb20iLCJmaGlyVXNlciI6IlBhdGllbnQvMTIzNCJ9.k_uqVL_uXo49ETrhSwaNXw0LYDadvt4LJuwrKh-0FJo';
+const practitionerUserAllAccess: string =
+    'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2ZXIiOjEsImp0aSI6IkFULjZhN2tuY1RDcHUxWDllbzJRaEgxel9XTFVLNFR5VjQzbl85STZrWk53UFkiLCJpc3MiOiJodHRwczovL2Rldi02NDYwNjExLm9rdGEuY29tL29hdXRoMi9kZWZhdWx0IiwiYXVkIjoiYXBpOi8vZGVmYXVsdCIsImlhdCI6MTYwMzExODEzOCwiZXhwIjoxNjAzMTIxNzM4LCJjaWQiOiIwb2E4bXVhektTeWs5Z1A1eTVkNSIsInVpZCI6IjAwdTg1b3p3ampXUmQxN1BCNWQ1Iiwic2NwIjpbImZoaXJVc2VyIiwib3BlbmlkIiwicHJvZmlsZSIsImxhdW5jaC9lbmNvdW50ZXIiLCJwYXRpZW50L1BhdGllbnQucmVhZCIsInBhdGllbnQvT2JzZXJ2YXRpb24ucmVhZCIsInVzZXIvKi4qIl0sInN1YiI6InNtYXlkYTQ0QGdtYWlsLmNvbSIsImZoaXJVc2VyIjoiUHJhY3RpdGlvbmVyLzEyMzQifQ.hUvJkUw108XDDw7n6KrRvKbzMWSfp84kt4-qqKCRXA4';
+const practitionerUserReadAccess: string =
+    'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2ZXIiOjEsImp0aSI6IkFULjZhN2tuY1RDcHUxWDllbzJRaEgxel9XTFVLNFR5VjQzbl85STZrWk53UFkiLCJpc3MiOiJodHRwczovL2Rldi02NDYwNjExLm9rdGEuY29tL29hdXRoMi9kZWZhdWx0IiwiYXVkIjoiYXBpOi8vZGVmYXVsdCIsImlhdCI6MTYwMzExODEzOCwiZXhwIjoxNjAzMTIxNzM4LCJjaWQiOiIwb2E4bXVhektTeWs5Z1A1eTVkNSIsInVpZCI6IjAwdTg1b3p3ampXUmQxN1BCNWQ1Iiwic2NwIjpbImZoaXJVc2VyIiwib3BlbmlkIiwicHJvZmlsZSIsImxhdW5jaC9lbmNvdW50ZXIiLCJwYXRpZW50L1BhdGllbnQucmVhZCIsInBhdGllbnQvT2JzZXJ2YXRpb24ucmVhZCIsInVzZXIvKi5yZWFkIl0sInN1YiI6InNtYXlkYTQ0QGdtYWlsLmNvbSIsImZoaXJVc2VyIjoiUHJhY3RpdGlvbmVyLzEyMzQifQ.85sfPi7_PoNS4zhUeoQWv-dvEUQSwga-h77DT6SXH_M';
 const manyWriteAccess: string =
     'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2ZXIiOjEsImp0aSI6IkFULjZhN2tuY1RDcHUxWDllbzJRaEgxel9XTFVLNFR5VjQzbl85STZrWk53UFkiLCJpc3MiOiJodHRwczovL2Rldi02NDYwNjExLm9rdGEuY29tL29hdXRoMi9kZWZhdWx0IiwiYXVkIjoiYXBpOi8vZGVmYXVsdCIsImlhdCI6MTYwMzExODEzOCwiZXhwIjoxNjAzMTIxNzM4LCJjaWQiOiIwb2E4bXVhektTeWs5Z1A1eTVkNSIsInVpZCI6IjAwdTg1b3p3ampXUmQxN1BCNWQ1Iiwic2NwIjpbImZoaXJVc2VyIiwib3BlbmlkIiwicHJvZmlsZSIsInVzZXIvUGF0aWVudC53cml0ZSIsInN5c3RlbS9PYnNlcnZhdGlvbi53cml0ZSIsInBhdGllbnQvKi53cml0ZSJdLCJzdWIiOiJzbWF5ZGE0NEBnbWFpbC5jb20iLCJmaGlyVXNlciI6IlBhdGllbnQvMTIzNCJ9.tRx8pf60I98vUJI7Q87sZkvI24ii6ADQ_jSw88Q42EY';
 const allSysAccess: string =
@@ -323,6 +327,76 @@ describe('verifyAccessToken; scopes are in an array', () => {
         }
         const identity = addScopeToUserIdentity(
             patientIdentityWithoutScopes,
+            (<VerifyAccessTokenRequest>request).accessToken,
+        );
+        return expect(authZHandler.verifyAccessToken(<VerifyAccessTokenRequest>request)).resolves.toEqual(identity);
+    });
+});
+
+describe('verifyAccessToken; System level export requests', () => {
+    const arrayScopesCases: (string | boolean | VerifyAccessTokenRequest)[][] = [
+        [
+            'Read and Write Access: initiate-export',
+            {
+                accessToken: practitionerUserAllAccess,
+                operation: 'read',
+                resourceType: '',
+                bulkDataAuth: { exportType: 'system', operation: 'initiate-export' },
+            },
+            true,
+        ],
+        [
+            'Read Access: initiate-export',
+            {
+                accessToken: practitionerUserReadAccess,
+                operation: 'read',
+                resourceType: '',
+                bulkDataAuth: { exportType: 'system', operation: 'initiate-export' },
+            },
+            true,
+        ],
+        [
+            'Read and Write Access: get-status-export',
+            {
+                accessToken: practitionerUserAllAccess,
+                operation: 'read',
+                resourceType: '',
+                bulkDataAuth: { exportType: 'system', operation: 'get-status-export' },
+            },
+            true,
+        ],
+        [
+            'Read and Write Access: cancel-export',
+            {
+                accessToken: practitionerUserAllAccess,
+                operation: 'read',
+                resourceType: '',
+                bulkDataAuth: { exportType: 'system', operation: 'cancel-export' },
+            },
+            true,
+        ],
+        [
+            'No User read access: initiate-export',
+            {
+                accessToken: manyReadAccess,
+                operation: 'read',
+                resourceType: '',
+                bulkDataAuth: { exportType: 'system', operation: 'cancel-export' },
+            },
+            false,
+        ],
+    ];
+
+    const authZHandler: SMARTHandler = new SMARTHandler(authZConfig, apiUrl, '4.0.1');
+    test.each(arrayScopesCases)('CASE: %p', (_firstArg, request, isValid) => {
+        mock.onGet(authZConfig.userInfoEndpoint).reply(200, practitionerIdentityWithoutScopes);
+        if (!isValid) {
+            return expect(authZHandler.verifyAccessToken(<VerifyAccessTokenRequest>request)).rejects.toThrowError(
+                UnauthorizedError,
+            );
+        }
+        const identity = addScopeToUserIdentity(
+            practitionerIdentityWithoutScopes,
             (<VerifyAccessTokenRequest>request).accessToken,
         );
         return expect(authZHandler.verifyAccessToken(<VerifyAccessTokenRequest>request)).resolves.toEqual(identity);
--- a/src/smartHandler.test.ts
+++ b/src/smartHandler.test.ts
@@ -12,6 +12,10 @@ import {
     UnauthorizedError,
     WriteRequestAuthorizedRequest,
     clone,
+    AccessBulkDataJobRequest,
+    AllowedResourceTypesForOperationRequest,
+    BASE_R4_RESOURCES,
+    AuthorizationBundleRequest,
 } from 'fhir-works-on-aws-interface';
 import { decode } from 'jsonwebtoken';
 import { SMARTHandler } from './smartHandler';
@@ -97,7 +101,7 @@ const authZConfig: SMARTConfig = {
 const apiUrl = 'https://fhir.server.com/dev/';
 const patientId = 'Patient/1234';
 const patientIdentityWithoutScopes = { [authZConfig.fhirUserClaimKey]: `${apiUrl}${patientId}` };
-const validPractitionerIdentity = { [authZConfig.fhirUserClaimKey]: `${apiUrl}Practitioner/1234` };
+const practitionerIdentityWithoutScopes = { [authZConfig.fhirUserClaimKey]: `${apiUrl}Practitioner/1234` };
 const validExternalPractitionerIdentity = { [authZConfig.fhirUserClaimKey]: `${apiUrl}test/Practitioner/1234` };
 
 const validPatient = {
@@ -600,7 +604,7 @@ describe('authorizeAndFilterReadResponse', () => {
         [
             'READ: Practitioner able to read Encounter',
             {
-                userIdentity: validPractitionerIdentity,
+                userIdentity: practitionerIdentityWithoutScopes,
                 operation: 'read',
                 readResponse: validPatientEncounter,
             },
@@ -610,7 +614,7 @@ describe('authorizeAndFilterReadResponse', () => {
         [
             'READ: Practitioner able to read unrelated Observation',
             {
-                userIdentity: validPractitionerIdentity,
+                userIdentity: practitionerIdentityWithoutScopes,
                 operation: 'read',
                 readResponse: validPatientObservation,
             },
@@ -690,7 +694,7 @@ describe('authorizeAndFilterReadResponse', () => {
         [
             'SEARCH: Practitioner able to search and get ALL results',
             {
-                userIdentity: validPractitionerIdentity,
+                userIdentity: practitionerIdentityWithoutScopes,
                 operation: 'search-type',
                 readResponse: searchAllEntitiesMatch,
             },
@@ -726,7 +730,7 @@ describe('isWriteRequestAuthorized', () => {
         [
             'Practitioner able to write Patient',
             {
-                userIdentity: validPractitionerIdentity,
+                userIdentity: practitionerIdentityWithoutScopes,
                 operation: 'vread',
                 resourceBody: validPatient,
             },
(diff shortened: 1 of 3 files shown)


src/smartHandler.ts
```diff
@@ -97,26 +109,104 @@ export class SMARTHandler implements Authorization {
...
+            // If the scope allows request.operation, get the resources allowed for that operation as defined by 'this.operationToAllowedResources'
+            if (validOperations.includes(request.operation)) {
+                let match = scope.match(SMARTHandler.LAUNCH_SCOPE_REGEX);
+                if (!match) {
+                    match = scope.match(SMARTHandler.CLINICAL_SCOPE_REGEX);
+                }
+                if (match !== null) {
+                    const { scopeType } = match.groups!;
+                    const allowedResourcesForScope = this.operationToAllowedResources[request.operation];
+                    if (scopeType === 'launch') {
```
nguyen102: @rsmayda Should we include `Encounter` and `Patient` as `allowedResources` if they're included in the launch scope? I'm leaning towards no, and therefore deleting the `if statement` here. However, I wanted to double check with you on this.

src/smartHandler.ts
```diff
@@ -38,18 +39,23 @@ export class SMARTHandler implements Authorization {
...
     private readonly version: number = 1.0;
 
     private readonly config: SMARTConfig;
 
     private readonly apiUrl: string;
 
-    constructor(config: SMARTConfig, apiUrl: string) {
+    private readonly operationToAllowedResources: Record<string, string[]>;
+
+    constructor(config: SMARTConfig, apiUrl: string, operationToAllowedResources: Record<string, string[]>) {
```
nguyen102: Example
```
operationToAllowedResources = { 'search-type': ['Patient', 'Observation', 'Medication']
```

src/smartHandler.ts
```diff
@@ -97,26 +109,104 @@ export class SMARTHandler implements Authorization {
...
+            }
         }
-        return response.data;
+
+        const userIdentity = clone(response.data);
+        userIdentity.scopes = scopes;
+        return userIdentity;
+    }
+
+    private getScopes(decoded: any): string[] {
```
rsmayda:
```suggestion
    private getScopes(scopes: string | string[]): string[] {
```

src/smartHandler.ts
rsmayda: Why is this needed? can we have a default to the Patient compartment? Do you see this being used for other operations i.e. write, read, etc?

src/smartHandler.ts
```diff
@@ -97,26 +109,104 @@ export class SMARTHandler implements Authorization {
...
+        if (this.config.scopeValueType === 'space' && typeof decoded[this.config.scopeKey] === 'string') {
+            return decoded[this.config.scopeKey].split(' ');
+        }
+        if (this.config.scopeValueType === 'array' && Array.isArray(decoded[this.config.scopeKey])) {
+            return decoded[this.config.scopeKey];
+        }
+        return [];
     }
 
     // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
```
rsmayda, line 139: remove

src/smartHandler.ts
```diff
@@ -97,26 +109,104 @@ export class SMARTHandler implements Authorization {
...
     async isAccessBulkDataJobAllowed(request: AccessBulkDataJobRequest): Promise<void> {
-        // TODO not supported for now
-        throw new UnauthorizedError('Bulk access is not allowed');
+        if (request.userIdentity.sub !== request.jobOwnerId) {
+            throw new UnauthorizedError('User does not have permission to access this Bulk Data Export job');
+        }
     }
 
     // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
```
rsmayda: remove

src/smartHandler.ts
```diff
@@ -97,26 +109,104 @@ export class SMARTHandler implements Authorization {
...
     async isBundleRequestAuthorized(request: AuthorizationBundleRequest): Promise<void> {
```
rsmayda, line 146: on the output of the bundle we will need to call; `authorizeAndFilterReadResponse` to ensure that the bundle can read are the responses

src/smartHandler.ts
```diff
@@ -97,26 +109,104 @@ export class SMARTHandler implements Authorization {
...
+        console.log('beginning of getAllowedResourceTypesForOperation');
+        const { scopes } = request.userIdentity;
+        let allowedResources: string[] = [];
+
+        for (let i = 0; i < scopes.length; i += 1) {
+            const scope = scopes[i];
+            const validOperations = this.getValidOperationsForScope(scope);
+            console.log('validOperations', validOperations);
+            // If the scope allows request.operation, get the resources allowed for that operation as defined by 'this.operationToAllowedResources'
+            if (validOperations.includes(request.operation)) {
```
rsmayda: This should already be verified by the verify access token method right?

src/smartHandler.ts
rsmayda: Yep same feeling

src/smartHandler.ts
```diff
@@ -97,26 +109,104 @@ export class SMARTHandler implements Authorization {
...
+
+        try {
+            await Promise.all(authWritePromises);
+        } catch (e) {
+            throw new UnauthorizedError('Bundle operation is not authorized');
+        }
     }
 
     // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
     async getAllowedResourceTypesForOperation(request: AllowedResourceTypesForOperationRequest): Promise<string[]> {
```
rsmayda, line 177: This is a confusing method name since it is only used for include/rev_include -- not sure if we have an alternative but it is confusing

src/smartHandler.ts
```diff
@@ -190,41 +280,66 @@ export class SMARTHandler implements Authorization {
...
+            );
+            if (validOperations.includes(operation)) return true;
+        }
+        return false;
+    }
+
+    private getValidOperationsForScope(
+        scope: string,
+        resourceType?: string,
+        isTransactionOperation: boolean = false,
```
rsmayda: should we generalize to be 'bundle' ie support same behavior for 'batch' or 'transaction'

src/smartHandler.ts
```diff
@@ -190,41 +280,66 @@ export class SMARTHandler implements Authorization {
         operation: TypeOperation | SystemOperation,
         resourceType?: string,
     ): boolean {
-        const { scopeRule } = this.config;
         for (let i = 0; i < scopes.length; i += 1) {
             const scope = scopes[i];
```
rsmayda, line 285: nit can probable remove this variable since it is just be used once

src/smartHandler.ts
```diff
@@ -190,41 +280,66 @@ export class SMARTHandler implements Authorization {
...
                 }
-                if (validOperations.includes(operation)) return true;
             }
         }
-        return false;
+        return validOperations;
     }
 
-    private getValidOperationsForScope(scopeType: ScopeType, accessType: string) {
+    private getValidOperationsForScopeGivenScopeRule(
```
rsmayda: I personally prefer the older name the `GivenScopeRule` doesn't add clarity to me - what do you think? or what about the old name is confusing?

src/smartHandler.ts
```diff
@@ -190,41 +280,66 @@ export class SMARTHandler implements Authorization {
...
+                    validOperations = this.getValidOperationsForScopeGivenScopeRule(<ScopeType>scopeType, '*');
+                }
+                if (resourceType) {
+                    if (scopeResourceType === '*' || scopeResourceType === resourceType) {
+                        validOperations = this.getValidOperationsForScopeGivenScopeRule(
+                            <ScopeType>scopeType,
+                            accessType,
+                        );
                     }
+                } else if (scopeResourceType === '*' || resourceType === undefined) {
```
rsmayda: hmmm I think you just turn this into an `else`?

for the use case below, should we authorize it? I almost think for system level operations they **need** to have scope resourceType to be '*'

example:
resourceType: null
scope: patient/Obervation.*
operation: search-system

src/smartHandler.ts
```diff
@@ -190,41 +280,66 @@ export class SMARTHandler implements Authorization {
...
+                if (['patient', 'encounter'].includes(launchType)) {
+                    validOperations = scopeRule[scopeType][<LaunchType>launchType];
+                } else if (!launchType) {
+                    validOperations = scopeRule[scopeType].launch;
+                }
+            } else if (['patient', 'user', 'system'].includes(scopeType)) {
+                const { scopeResourceType, accessType } = match.groups!;
+                // transaction operations do not have a resourceType
+                // entries in a transaction can be either 'read' or 'write' therefore we need to check if 'transaction' is allowed in the scope rules for 'read' or 'write'
+                if (isTransactionOperation) {
```
rsmayda: I think if it is a transaction then the scope's resource type must be a '*'... thoughts?

rsmayda (commented):
> scopes is needed later on by isBundleRequestAuthorized and getAllowedResourceTypesForOperation. These two methods are called in the handler classes of the router package.

I'd like to not pass the scopes further on do we really need the scopes for Bundles/system level operations?

src/smartHandler.ts
nguyen102: After looking at the code some more and thinking through more cases, I think we don't need `operationToAllowedResources` anymore. Based on the scopes provided, we can determine the `allowedResources`.

example:
After getting the validOperations for `user/Patient.read` from `ScopeRule`, `allowedResources` would equal `['Patient']`

After getting the validOperations for `user/*.read` from `ScopeRule`, `allowedResources` would equal `[All base R4 resources or STU3 resources]`

src/smartHandler.ts
nguyen102: Isn't decoded an object map? 
Example usage of decoded
`decoded[this.config.scopeKey]`

`this.config.scopeKey` is a string. 

Should the type signature for decoded be
`decoded: Record<string, any>`

src/smartHandler.ts
nguyen102: done

src/smartHandler.ts
nguyen102: sounds good, i'll make a note of that

src/smartHandler.ts
nguyen102: done

src/smartHandler.ts
nguyen102: Hmm, that's a good point for `search-system`. I'll revert this if branch back to `else if (scopeResourceType === '*')`. 

We're already handling the `transaction` case with `if (isTransactionOperation) `. For `transaction`, a request can still succeed if the entries within the `transaction` are within the user's scopes. For example, if the transaction is only to read a Patient, then only `/Patient.read` is required.

src/smartHandler.ts
nguyen102: For transaction, a request can still succeed if the entries within the transaction are within the user's scopes. For example, if the transaction is only to read a Patient, then only /Patient.read is required.

src/smartHandler.ts
nguyen102: I pulled out the `if statement` inside of `areScopesSufficient` and put it in a new method. I felt like the proper name for that would be `getValidOperationsForScope`. And then this method on line 339 would become `getValidOperationsForScopeGivenScopeRule` as it's now a helper method for the new `getValidOperationsForScope`

src/smartHandler.ts
nguyen102: yep, that works. Making the change.

src/smartHandler.ts
rsmayda: What I am thinking is getScopes doesn't need the entire decoded just the scopes field

src/smartHandler.ts
```diff
@@ -66,12 +75,7 @@ export class SMARTHandler implements Authorization {
         }
 
         // verify scope
-        let scopes: string[] = [];
-        if (this.config.scopeValueType === 'space' && typeof decoded[this.config.scopeKey] === 'string') {
-            scopes = decoded[this.config.scopeKey].split(' ');
-        } else if (this.config.scopeValueType === 'array' && Array.isArray(decoded[this.config.scopeKey])) {
-            scopes = decoded[this.config.scopeKey];
-        }
+        const scopes = this.getScopes(decoded);
```
rsmayda: I am thinking this will be -> `decoded[config.scopeKey]`

src/smartHandler.ts
```diff
@@ -97,26 +102,87 @@ export class SMARTHandler implements Authorization {
...
     async isBundleRequestAuthorized(request: AuthorizationBundleRequest): Promise<void> {
-        // TODO not supported for now
-        throw new UnauthorizedError('Bundle operation is not authorzied');
+        const { scopes } = request.userIdentity;
+        request.requests.forEach((req: BatchReadWriteRequest) => {
+            if (!this.areScopesSufficient(scopes, req.operation, req.resourceType)) {
+                console.error(
+                    `User supplied scopes are insufficient\nscopes: ${scopes}\noperation: ${req.operation}\nresourceType: ${req.resourceType}`,
+                );
+                throw new UnauthorizedError('Bundle operation is not authorized');
```
rsmayda: Change it to be more specific
```suggestion
                throw new UnauthorizedError('An operation within the bundle is not authorized');

```

src/smartHandler.ts
```diff
@@ -97,26 +102,87 @@ export class SMARTHandler implements Authorization {
...
+        const authWritePromises = request.requests.map(req => {
+            if (['create', 'update', 'patch', 'delete'].includes(req.operation)) {
+                return this.isWriteRequestAuthorized(<WriteRequestAuthorizedRequest>{
+                    userIdentity: request.userIdentity,
+                    operation: req.operation,
+                    resourceBody: req.resource,
+                });
+            }
+            return Promise.resolve();
+        });
```
rsmayda: Could you combine these two loops into one? then each loop check both conditions?

src/smartHandler.ts
```diff
@@ -97,26 +102,87 @@ export class SMARTHandler implements Authorization {
...
+            if (validOperations.includes(request.operation)) {
+                const match = scope.match(SMARTHandler.CLINICAL_SCOPE_REGEX);
+                if (match !== null) {
+                    const { scopeType } = match.groups!;
+                    const allowedResourcesForScope: string[] =
+                        this.fhirVersion === '4.0.1' ? BASE_R4_RESOURCES : BASE_STU3_RESOURCES;
+                    if (['patient', 'user', 'system'].includes(scopeType)) {
+                        const { scopeResourceType } = match.groups!;
+                        if (scopeResourceType === '*') {
+                            allowedResources = allowedResources.concat(allowedResourcesForScope);
```
rsmayda: If we reach this case shouldn't we return all resource type? i.e. you can't add more resource types to all?

src/smartHandler.ts
```diff
@@ -97,26 +102,87 @@ export class SMARTHandler implements Authorization {
...
+                        } else if (allowedResourcesForScope.includes(scopeResourceType)) {
+                            allowedResources = allowedResources.concat(scopeResourceType);
+                        }
+                    }
+                }
+            }
+        });
+        allowedResources = [...new Set(allowedResources)];
```
rsmayda, line 208: I don't know if we will need this if we return quickly in the case above

src/smartHandler.ts
```diff
@@ -190,41 +256,69 @@ export class SMARTHandler implements Authorization {
...
+                else if (['search-system', 'history-system'].includes(reqOperation)) {
+                    if (scopeResourceType === '*') {
+                        validOperations = this.getValidOperationsForScopeGivenScopeRule(
+                            <ScopeType>scopeType,
+                            accessType,
+                        );
                     }
+                } else {
+                    validOperations = this.getValidOperationsForScopeGivenScopeRule(<ScopeType>scopeType, accessType);
                 }
```
rsmayda, line 312: What is the use-case for this else?

if the scope is not launch, patient, user, system, what is it? What should we return

rsmayda: I still struggle to see why we want add this

if
1. 'patient', 'user', 'system' scope 
1. the request resource is null 
1. it is not a system history/search

If there is no resource defined shouldn't we require the scopeResourceType =='*'? If this is for batch/transaction could we add an edge-case that way instead of search?

nguyen102: That's a good question. It's structured this way to allow for two different use case. The first case is the `batch/transaction` use case.

The second case is when `getValidOperationsForScope` is called by [getAllowedResourceTypesForOperation](https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/ddd77d0163f2b3b14eab41e83f92607f3495bc2c/src/smartHandler.ts#L165). In this second use case we have the scopes and the operation as provided by `AllowedResourceTypesForOperationRequest`. However, we do not have a `resourceType`. Therefore it could be possible where `scope = user/Patient.read` and `operation = search-type`. In this scenario we would still want to get the `validOperations` for that `scope` and `operation`.

src/smartHandler.ts
nguyen102: updated

src/smartHandler.ts
nguyen102: updated

src/smartHandler.ts
nguyen102: updated

src/smartHandler.ts
nguyen102: done

src/smartHandler.ts
nguyen102: good idea. Updated

src/smartHandler.ts
nguyen102: If there are three scopes as shown below
`['user/Patient.read', 'patient/Observation.read', 'user/Observation.read']`
we would want to only return `[Patient, Observation]` instead of `[Patient, Observation, Observation]`

src/smartHandler.ts
```diff
@@ -133,35 +133,36 @@ export class SMARTHandler implements Authorization {
...
     async isBundleRequestAuthorized(request: AuthorizationBundleRequest): Promise<void> {
         const { scopes } = request.userIdentity;
+        const authWritePromises: Promise<void>[] = [];
         request.requests.forEach((req: BatchReadWriteRequest) => {
             if (!this.areScopesSufficient(scopes, req.operation, req.resourceType)) {
                 console.error(
                     `User supplied scopes are insufficient\nscopes: ${scopes}\noperation: ${req.operation}\nresourceType: ${req.resourceType}`,
                 );
-                throw new UnauthorizedError('Bundle operation is not authorized');
+                throw new UnauthorizedError('An operation with the Bundle is not authorized');
```
rsmayda: Since we are throwing error here do we have to be concerned for the open promises? I don't think so; But want to verify https://mrcoles.com/javascript-promises-and-errors/

src/smartHandler.ts
rsmayda: Interesting Should we change `AllowedResourceTypesForOperationRequest` to include the resourceType?

rsmayda: Looks like it already is in requesT: https://github.com/awslabs/fhir-works-on-aws-authz-smart/pull/7/files#diff-7f9bca42e8ef8f0cc4b82957be62aa94575c0886947d59b067cd8f1eaea5ba6eR166

src/smartHandler.ts
```diff
@@ -97,26 +102,89 @@ export class SMARTHandler implements Authorization {
...
+                        }
+                    }
+                }
+            }
+        }
+        allowedResources = [...new Set(allowedResources)];
+        return allowedResources;
     }
 
     async authorizeAndFilterReadResponse(request: ReadResponseAuthorizedRequest): Promise<any> {
```
rsmayda, line 212: Bundles are currently not going through this method; this is a gap could we add a SIM task for this? we need to add this method call to Routing package and add bundle logic here

src/smartHandler.ts
rsmayda: For a concrete use-case that I am not sure we are covering say we have rule as:
`{patient:{read:['transaction', 'read']}}`
and a bundle request:
`POST ehr.com/`
with scopes:
`patient/Patient.read`

For this settings we would want to allow the first pass from `verifyaccesstoken` method (checking if the scope gives transaction permission) then in isBundleAuthorized we want to go 1 by 1 to verify if the contents are permitted

rsmayda: I guess is there any concern just to make this:
```suggestion
                } else if (['bundle',...].includes(reqOperation)) {
                    validOperations = this.getValidOperationsForScopeGivenScopeRule(<ScopeType>scopeType, accessType);
                }

```

src/smartHandler.ts
nguyen102: I added this requirement to the current FHIR-419 task. And I'm in the process of creating a PR for this.

src/smartHandler.test.ts
```diff
@@ -742,3 +760,171 @@ describe('isWriteRequestAuthorized', () => {
...
+        if (!isAuthorized) {
+            await expect(authZHandler.isBundleRequestAuthorized(request)).rejects.toThrowError(
+                new UnauthorizedError('An entry within the Bundle is not authorized'),
+            );
+        } else {
+            await expect(authZHandler.isBundleRequestAuthorized(request)).resolves.not.toThrow();
+        }
+    });
+
+    test('Bundle request with first Bundle entry passing and second Bundle entry failing', async () => {
```
nguyen102, line 929: Added this test to ensure that we handle "open promises" correctly.

src/smartHandler.ts
nguyen102: Added a test for this.

src/smartHandler.ts
```diff
@@ -97,26 +111,94 @@ export class SMARTHandler implements Authorization {
...
-    // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
     async isBundleRequestAuthorized(request: AuthorizationBundleRequest): Promise<void> {
-        // TODO not supported for now
-        throw new UnauthorizedError('Bundle operation is not authorzied');
+        const { scopes } = request.userIdentity;
+        const authWritePromises: Promise<void>[] = [];
+        request.requests.forEach((req: BatchReadWriteRequest) => {
+            if (!this.areScopesSufficient(scopes, req.operation, req.resourceType)) {
+                console.error(
+                    `User supplied scopes are insufficient\nscopes: ${scopes}\noperation: ${req.operation}\nresourceType: ${req.resourceType}`,
```
carvantes: IMO adding newlines in logs just makes them a lot harder to parse for the very little benefit of them looking prettier

src/smartHandler.ts
carvantes: Adding a test does not fix the issue.

There is an actual risk here of unhandled promise rejections.

you need to refactor to avoid dangling promises(or "open promises"). The easiest would be to just do the sync `areScopesSufficient` check first for ALL requests and after all sync checks pass do the async `isWriteRequestAuthorized` for all requests.

src/smartHandler.ts
```diff
@@ -97,26 +111,94 @@ export class SMARTHandler implements Authorization {
...
+                                return allowedResourcesForScope;
+                            }
+                            if (allowedResourcesForScope.includes(scopeResourceType)) {
+                                allowedResources = allowedResources.concat(scopeResourceType);
+                            }
+                        }
+                    }
+                }
+                // eslint-disable-next-line no-empty
+            } catch (e) {}
```
carvantes: Add a comment on the empty catch explaining why it's ok to swallow the exception here. That's better than disabling the eslint rule.

src/smartHandler.ts
nguyen102: @carvantes what is the impact of an unhandled promise rejection? Does that cause a memory leak? In this case if `areScopesSufficient` fails, we will return an error and the requester will get a response back, which is the expected behavior. However, if this code causes a memory leak, then let's be safe and code it the way you suggested.

carvantes: 1. It's messy code. This particular case is rather simple, but it can be hard to understand the code flow when leaving dangling promises with bigger side effects. 
2. Right now it is only a warning, but in upcoming versions of node js it will crash the process:

```
(node:90725) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:90725) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

src/smartHandler.ts
nguyen102: updated

src/smartHandler.ts
nguyen102: Rewriting the logging to this format to maintain readability, while making the logs easier to parse.
```
console.error('User supplied scopes are insufficient', {
                scopes,
                operation: request.operation,
                resourceType: request.resourceType,
            });
```

src/smartHandler.ts
nguyen102: Thanks for the additional info @carvantes . To be safe, I'll revert the code back to using two for loops.

rsmayda: @carvantes Yep now I remember you call this out before too

src/smartHandler.ts
```diff
@@ -133,54 +144,63 @@ export class SMARTHandler implements Authorization {
...
+                                return allowedResourcesForScope;
+                            }
+                            if (allowedResourcesForScope.includes(scopeResourceType)) {
+                                allowedResources = allowedResources.concat(scopeResourceType);
+                            }
                         }
                     }
                 }
+            } catch (e) {
+                // We only get allowedResourceTypes for ClinicalSmartScope
```
rsmayda:
```suggestion
                // Caused by casting to ClinicalSmartScope error; We only get allowedResourceTypes for ClinicalSmartScope

```

rsmayda (dismissed):
Thanks for the hard work Tim everything looks good

src/smartHandler.ts
nguyen102: Ah this error is actually caused by trying to convert non-SmartScope to SmartScope, for example converting scope 'openid' or 'profile'.

I'll update the comment to reflect that.

rsmayda (dismissed):

src/smartHandler.ts
rsmayda: +1 thanks

src/smartHandler.ts
```diff
@@ -278,15 +279,39 @@ export class SMARTHandler implements Authorization {
...
-            if (validOperations.includes(operation)) return true;
+            let smartScope: SmartScope | undefined;
+            try {
+                smartScope = this.convertScopeToSmartScope(scope);
+            } catch (e) {
+                // Caused by trying to convert non-SmartScope to SmartScope, for example converting scope 'openid' or 'profile'
+                // We don't need to check non-SmartScope
+            }
+            if (smartScope) {
+                if (bulkDataAuth) {
```
rsmayda, line 294: We could move this into a new method `areScopesSufficientForBulkDataAcess`, thoughts?

src/smartHandler.ts
```diff
@@ -278,15 +279,39 @@ export class SMARTHandler implements Authorization {
...
+                    const bulkDataRequestHasCorrectScope =
+                        bulkDataAuth.exportType === 'system' && // As of 12/9/20 we only support System Level export
+                        smartScope.scopeType === 'user' &&
+                        smartScope.resourceType === '*' &&
+                        ['*', 'read'].includes(smartScope.accessType);
```
rsmayda: I think we are missing the check to see if the operation is in the rules or not? ie we are hard coding that user/*.read has export access will in the rule that may not be the case

rsmayda (approved):

PR #8, opened 2020-12-07
Reducing access modifiers to protected

emilhdiaz:
Description of changes:

This change reduces the access modifiers of certain methods and properties in the `SMARTHandler` such that subclassing by extending this class can more easily override method implementations without running into errors about unaccessible private methods/properties. 

### DIFF
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -36,13 +36,13 @@ export class SMARTHandler implements Authorization {
         'history-system',
     ];
 
-    private readonly typesWithWriteAccess: IdentityType[] = ['Practitioner'];
+    protected readonly typesWithWriteAccess: IdentityType[] = ['Practitioner'];
 
-    private readonly version: number = 1.0;
+    protected readonly version: number = 1.0;
 
-    private readonly config: SMARTConfig;
+    protected readonly config: SMARTConfig;
 
-    private readonly apiUrl: string;
+    protected readonly apiUrl: string;
 
     constructor(config: SMARTConfig, apiUrl: string) {
         if (config.version !== this.version) {
@@ -185,7 +185,7 @@ export class SMARTHandler implements Authorization {
         throw new UnauthorizedError("Requester's identity is in the incorrect format");
     }
 
-    private areScopesSufficient(
+    protected areScopesSufficient(
         scopes: string[],
         operation: TypeOperation | SystemOperation,
         resourceType?: string,


emilhdiaz:
@rsmayda Below is an excerpt from our subclass for reference: 

```
export default class AbacusSMARTHandler extends SMARTHandler {
    private static INTERNAL_META_KEY = "__internalMeta"
    private readonly dataService: Persistence
    private userinfo: KeyValueMap;

    constructor(config: SMARTConfig, dataService: Persistence) {
        super(config, '');
        this.dataService = dataService
        this.userinfo = {}
    }

    async verifyAccessToken(request: VerifyAccessTokenRequest): Promise<KeyValueMap> {
        // The access_token will be verified by hitting the authZUserInfoUrl (token introspection)
        // Decoding first to determine if it passes scope & claims check first
        const decoded = decode(request.accessToken, { json: true }) ?? {};
        const { aud, iss } = decoded;
        const audArray = Array.isArray(aud) ? aud : [aud];
        const scopes = this.extractScopes(decoded);

        // verify aud
        if ( !audArray.includes(this.config.expectedAudValue) ) {
            console.error(
                `aud claim did not include the expected audience -\n` +
                `aud: ${audArray.join(',')}\n` +
                `expected: ${this.config.expectedAudValue}`
            );
            throw new UnauthorizedError('Error validating the validity of the access_token');
        }

        // verify iss
        if ( this.config.expectedIssValue !== iss ) {
            console.error(
                `iss claim does not match the expected issuer -\n` +
                `iss: ${iss}\n` +
                `expected: ${this.config.expectedIssValue}`
            );
            throw new UnauthorizedError('Error validating the validity of the access_token');
        }

        // verify scopes
        if (!this.areScopesSufficient(scopes, request.operation, request.resourceType)) {
            console.error(
                `User supplied scopes are insufficient -\n` +
                `scopes: ${scopes}\n` +
                `operation: ${request.operation}\n` +
                `resourceType: ${request.resourceType}`
            );
            throw new UnauthorizedError('User does not have permission for requested operation');
        }

        // lookup userinfo
        await this.lookupUserinfo(request.accessToken);

        // verify fhirUser claim
        if (!this.userinfo[this.config.fhirUserClaimKey]) {
            console.error(`userinfo from Auth Server was missing the ${this.config.fhirUserClaimKey} claim`);
            throw new UnauthorizedError("Cannot determine requester's identity");
        }

        // verify the fhirMemberIDs claim
        if (!this.userinfo[this.config.fhirMemberIDsClaimKey]) {
            console.error(`userinfo from Auth Server was missing the ${this.config.fhirMemberIDsClaimKey} claim`);
            throw new UnauthorizedError("Cannot determine requester's identity");
        }

        return this.userinfo
    }

        const { operation, readResponse } = request;
        const { resourceType, id } = readResponse;
        const stripInternalMeta = (resource: any) => {
            delete resource[AbacusSMARTHandler.INTERNAL_META_KEY]
            return resource
        }

        // if searching/listing, filter out unauthorized records
        if (SMARTHandler.SEARCH_OPERATIONS.includes(operation)) {
            const entries = (readResponse.entry ?? [])
                .filterAsync((entry: { resource: any }) => this.hasMatchingContext(entry.resource))
                .map((entry: { resource: any }) => stripInternalMeta(entry.resource));
            return { ...readResponse, entry: entries, total: entries.length };
        }

        // if direct access, reject unauthorized access
        if ( !await this.hasMatchingContext(readResponse) ) {
            console.error(
                `The contexts for requested resource does not matched the allowed fhirMemberIDs for this user -\n` +
                `resource: ${resourceType}:${id}\n` +
                `contexts: ${JSON.stringify(readResponse[AbacusSMARTHandler.INTERNAL_META_KEY].contexts)}\n` +
                `fhirMemberIDs: ${this.userinfo[this.config.fhirMemberIDsClaimKey]}`
            );
            throw new UnauthorizedError("You are not authorized to access this resource");
        }
        return { ...stripInternalMeta(readResponse) }
    }

    // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
    async isWriteRequestAuthorized(request: WriteRequestAuthorizedRequest): Promise<void> {
        // TODO not supported for now
        throw new UnauthorizedError('User does not have permission for requested operation');
    }

    private async hasMatchingContext(resource: any): Promise<boolean> {
        const fhirMemberIDs = this.userinfo[this.config.fhirMemberIDsClaimKey]
        const internalMeta: InternalMeta = resource[AbacusSMARTHandler.INTERNAL_META_KEY] ?? {}
        const identifiers: Identifier[] = (internalMeta.contexts ?? [])
            .filter((context: Context) => context.type == ContextType.PATIENT)
            .flatMap((context: Context) => context.identifiers ?? [])

        for ( const identifier of identifiers ) {
            switch (identifier.type) {
                // Check for a direct match in the context
                case IdentifierType.MEMBER_ID: {
                    if ( fhirMemberIDs.includes(identifier.value) ) return true
                    break;
                }

                // Check for an indirect match through the Patient record in context
                case IdentifierType.PRIMARY: {
                    // Follow the literal link to the Patient record
                    const patientResource: any = await this.dataService.readResource({
                        resourceType: ContextType.PATIENT,
                        id: identifier.value
                    });

                    // Extract the MemberID identifiers of the Patient
                    const patientIdentifiers: string[] = (patientResource.resource.identifier ?? [])
                        .filter((identifier: KeyValueMap) => identifier.type == IdentifierType.MEMBER_ID)
                        .map((identifier: KeyValueMap) => identifier.value);

                    if ( patientIdentifiers.filter(identifier => fhirMemberIDs.includes(identifier)).length > 0 ) return true
                    break
                }
            }
        }

        return false;
    }

    private async lookupUserinfo(accessToken: string): Promise<void> {
        this.userinfo = {}
        try {
            const response = await axios.get(this.config.userInfoEndpoint, {
                headers: { Authorization: `Bearer ${accessToken}` },
            });
            if ( response.data ) {
                this.userinfo = response.data
            }
        } catch (e) {
            console.error('Error retrieving userinfo from Auth Server', e);
        }
    }

    private extractScopes(accessToken: KeyValueMap): string[] {
        let scopes: string[] = [];

        if (this.config.scopeValueType === 'space' && typeof accessToken[this.config.scopeKey] === 'string') {
            scopes = accessToken[this.config.scopeKey].split(' ');
        }
        else if (this.config.scopeValueType === 'array' && Array.isArray(accessToken[this.config.scopeKey])) {
            scopes = accessToken[this.config.scopeKey];
        }

        return scopes
    }
}
```

src/smartHandler.ts
```diff
@@ -36,13 +36,13 @@ export class SMARTHandler implements Authorization {
         'history-system',
     ];
```
rsmayda, line 38: For your provided example: https://github.com/awslabs/fhir-works-on-aws-authz-smart/pull/8#issuecomment-740209784

With inheritance you a can use the super methods ie: `super.verifyAccessToken(request);` or just not include them at all and have the parent class handle the method. ie you may want to use our `isWriteRequestAuthorized` instead of yours

Just wanted to verify `isWriteRequestAuthorized` is there on purpose :)

rsmayda (commented):
Looks good to me @nguyen102 - can you review too

src/smartHandler.ts
emilhdiaz: That's a great call out. So taking each overridden method at a time: 

**verifyAccessToken**
I changed the structure of what we log while performing the token checks, hence why we don't just call into `super.verifyAccessToken`. These changes could be candidates for an open source contribution if you're interested. 

**authorizeAndFilterReadResponse**
Our implementation here is different as we're using the internalMeta contents

**isWriteRequestAuthorized**
Our implementation of the `fhirUser` claim in the auth server doesn't conform to the `FHIR_USER_REGEX` format. Currently, we aren't able to generate `fhirUser` claims with UUIDs from the resource table, as the Auth Server has no access to this information. So this was mostly to side-step the call to `this.getFhirUser(request.userIdentity)` in the parent class. So yes intentional, but also likely temporary until we solve for generating proper `fhirUser` claims.

rsmayda:
@emilhdiaz We had some internal dialogue of what we would rather have our customer's do: **inheritance** or **composition**? We decided on **composition** as it reduces the risk of our implementation breaking yours (or more generally our customers), it will also allow us to radically change our implementation without hurting our customers. In SMART there is a good deal of code that could be shared so what we plan to do is create a `SMARTUtility` class that will provide all of that _shared_ functionality in a shared class. 

What that means is your implementation will implement `Authorization` without extending our `SMARTHandler`. But in your constructor you could create a `SMARTUtility` that can help in the implementation of your `Authorization`

Happy to hop on a call/slack to discuss more!

emilhdiaz:
That works as well! Feel free to use any of the sample code I posted above if that's of interest. Thanks!

rsmayda:
Hey Emil we pulled out the generic helpers: https://github.com/awslabs/fhir-works-on-aws-authz-smart/commit/90eb0c0418352e1a0a7aa37c4d5f0cb796235cef

Thanks for working through this with us

You may want to wait to pull from mainline until we merge in https://github.com/awslabs/fhir-works-on-aws-authz-smart/pull/13

PR #9, opened 2020-12-15
feat: Add support for GetSearchFilterBasedOnIdentity

nguyen102:
Description of changes:
* Add support for `GetSearchFilterBasedOnIdentity`

### DIFF
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -149,6 +149,7 @@ export class SMARTHandler implements Authorization {
         const fhirUser = this.getFhirUser(request.userIdentity);
         const { hostname, resourceType, id } = fhirUser;
 
+        // Create a SearchFilter to limit access to only resources that are referring to the requesting user
         if (resourceType !== 'Practitioner') {
             const searchFilter: SearchFilter = {
                 key: '_reference',
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -145,7 +145,7 @@ export class SMARTHandler implements Authorization {
         }
     }
 
-    async getSearchFilterBasedOnIdentity(request: GetSearchFilterBasedOnIdentity): Promise<SearchFilter[]> {
+    async getSearchFilterBasedOnIdentity(request: GetSearchFilterBasedOnIdentityRequest): Promise<SearchFilter[]> {
         const fhirUser = this.getFhirUser(request.userIdentity);
         const { hostname, resourceType, id } = fhirUser;
 
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -149,7 +149,7 @@ export class SMARTHandler implements Authorization {
         const fhirUser = this.getFhirUser(request.userIdentity);
         const { hostname, resourceType, id } = fhirUser;
 
-        if (resourceType !== 'Practitioner' && ['search-type', 'search-system'].includes(request.operation)) {
+        if (resourceType !== 'Practitioner') {
             return [
                 {
                     key: '_reference',
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -21,7 +21,7 @@ import {
     FhirVersion,
     BASE_STU3_RESOURCES,
     BulkDataAuth,
-    GetSearchFilterBasedOnIdentity,
+    GetSearchFilterBasedOnIdentityRequest,
     SearchFilter,
 } from 'fhir-works-on-aws-interface';
 import axios from 'axios';
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -156,7 +156,7 @@ export class SMARTHandler implements Authorization {
                         key: '_reference',
                         value: [`${hostname}${resourceType}/${id}`],
                         operator: '==',
-                        logicalOperator: 'N/A',
+                        logicalOperator: 'AND', // logicalOperator can be either 'AND' or 'OR' since value is an array of one string
                     },
                 ];
             }
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -21,6 +21,8 @@ import {
     FhirVersion,
     BASE_STU3_RESOURCES,
     BulkDataAuth,
+    GetSearchFilterBasedOnIdentity,
+    SearchFilter,
 } from 'fhir-works-on-aws-interface';
 import axios from 'axios';
 import {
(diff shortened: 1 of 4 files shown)


src/smartHandler.ts
```diff
@@ -143,6 +145,24 @@ export class SMARTHandler implements Authorization {
...
+        const fhirUser = this.getFhirUser(request.userIdentity);
+        const { hostname, resourceType, id } = fhirUser;
+
+        if (resourceType !== 'Practitioner' && ['search-type', 'search-system'].includes(request.operation)) {
+            return [
+                {
+                    key: '_reference',
+                    value: [`${resourceType}/${id}`, `${hostname}${resourceType}/${id}`],
+                    operator: '==',
+                    logicalOperator: 'OR',
```
nguyen102: @rsmayda What are your thoughts on adding `logicalOperator` to the interface?

This allows us to filter by the Patient's full url *or* the Patient's relative url.

The new interface would look like this
```
export interface SearchFilter {
    key: string;
    value: string[];
    operator: '==' | '!=' | '>' | '<' | '>=' | '<=';
    logicalOperator: 'AND' | 'OR' | 'N/A';
}
```

nguyen102:
Blocked on merging of interface into mainline first.

src/smartHandler.ts
```diff
@@ -143,6 +145,24 @@ export class SMARTHandler implements Authorization {
         }
     }
 
+    async getSearchFilterBasedOnIdentity(request: GetSearchFilterBasedOnIdentityRequest): Promise<SearchFilter[]> {
```
nguyen102, line 148: @rsmayda  For `GetSearchFilterBasedOnIdentityRequest` would this be the interface
```
export interface GetSearchFilterBasedOnIdentity {
    userIdentity: KeyValueMap;
    operation: 'search-type' | 'search-system';
}
```

Particularly operations, should it be just `search-type` and `search-system`?

rsmayda: Sounds good we should expand it to be history-type and history-system too

src/smartHandler.ts
```diff
@@ -143,6 +145,24 @@ export class SMARTHandler implements Authorization {
         }
     }
 
+    async getSearchFilterBasedOnIdentity(request: GetSearchFilterBasedOnIdentityRequest): Promise<SearchFilter[]> {
+        const fhirUser = this.getFhirUser(request.userIdentity);
+        const { hostname, resourceType, id } = fhirUser;
+
+        if (resourceType !== 'Practitioner' && ['search-type', 'search-system'].includes(request.operation)) {
```
rsmayda: Add `history`

src/smartHandler.ts
```diff
@@ -143,6 +145,24 @@ export class SMARTHandler implements Authorization {
...
 
+    async getSearchFilterBasedOnIdentity(request: GetSearchFilterBasedOnIdentityRequest): Promise<SearchFilter[]> {
+        const fhirUser = this.getFhirUser(request.userIdentity);
+        const { hostname, resourceType, id } = fhirUser;
+
+        if (resourceType !== 'Practitioner' && ['search-type', 'search-system'].includes(request.operation)) {
+            return [
+                {
+                    key: '_reference',
+                    value: [`${resourceType}/${id}`, `${hostname}${resourceType}/${id}`],
```
rsmayda: Out of curiosity what happens if `hostname` is null? there could be a use-case when `fhirUser` is relative and not literal i.e. Patient/1234 vs api.com/Patient/1234

src/smartHandler.ts
rsmayda: +1 sounds good it adds a bit of complexity on the Search side but its worthwhile

src/smartHandler.ts
nguyen102: Noted. I'm thinking we have the different `operations` listed as part of the `interface`. Then we won't have to do the check for operations here. The interface would look like this.
```
export interface GetSearchFilterBasedOnIdentityRequest {
    userIdentity: KeyValueMap;
    operation: 'search-type' | 'search-system' | 'history-type' | 'history-system';
}
```

src/smartHandler.ts
nguyen102: It would fail before it gets to here.
To get the fhirUser, we call this [method](https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/aeb89dab7dba66f3061345a461517b13d63f4395/src/smartHandler.ts#L287).

If the `fhirUserValue` is not a fullUrl we would throw an error. `getFhirUser` method is also used in `verifyAccessToken`, so if the `fhirUser` is not a full url, we would fail at `verifyAccessToken` first.

src/smartHandler.ts
nguyen102: Updated

src/smartHandler.ts
rsmayda: got it -- that made me think of the concern I had -- What if the FHIR user is not of the same host

for example:
* hostname == `fhir.com`
* fhirUser == `api.com/Patient/1234`

We wouldn't want to search the relative

nguyen102: Good point. I'll update the code to compare the `hostname`  to `fhirUser.hostname`. If the two hostnames match we'll filter by the `fullUrl` and `relativeUrl`. If the two hostnames doesn't match, we'll only filter by `fullUrl`.

src/smartHandler.ts
```diff
@@ -143,6 +145,34 @@ export class SMARTHandler implements Authorization {
...
+        const { hostname, resourceType, id } = fhirUser;
+
+        if (resourceType !== 'Practitioner') {
+            if (hostname !== this.apiUrl) {
+                return [
+                    {
+                        key: '_reference',
+                        value: [`${hostname}${resourceType}/${id}`],
+                        operator: '==',
+                        logicalOperator: 'N/A',
```
rsmayda: I am not the biggest fan of having `N/A` as a logical operator; I feel like it should just be optional instead (separate PR though)

nguyen102: Updated to use the new interface from this PR.
https://github.com/awslabs/fhir-works-on-aws-interface/pull/37

src/smartHandler.ts
```diff
@@ -143,6 +145,34 @@ export class SMARTHandler implements Authorization {
...
+        const { hostname, resourceType, id } = fhirUser;
+
+        if (resourceType !== 'Practitioner') {
+            if (hostname !== this.apiUrl) {
+                return [
+                    {
+                        key: '_reference',
+                        value: [`${hostname}${resourceType}/${id}`],
+                        comparisonOperator: '==',
+                        logicalOperator: 'OR', // logicalOperator can be either 'AND' or 'OR' since value is an array of one string
```
rsmayda: Nit: I'd prefer to make this `AND`

nguyen102: @rsmayda Would you be ok with keeping this as an `OR`. I'm thinking of this as an `OR` to keep it consistent with line 168 below. Functionally it'll work the same whether it's an `OR` or an `AND`

src/smartHandler.ts
```diff
@@ -143,6 +145,26 @@ export class SMARTHandler implements Authorization {
         }
     }
 
+    async getSearchFilterBasedOnIdentity(request: GetSearchFilterBasedOnIdentityRequest): Promise<SearchFilter[]> {
+        const fhirUser = this.getFhirUser(request.userIdentity);
+        const { hostname, resourceType, id } = fhirUser;
+
```
rsmayda, line 151: Add minor comment something like:
```suggestion

// Create a SearchFilter to limit access to only resources that are referenced to the requesting user
```

nguyen102: updated

rsmayda (approved):

PR #11, opened 2021-01-04
feat: Add Smart handler helper

nguyen102:
Description of changes:
This PR moves all of the helper methods in `smartHandler.ts` out into 2 files: `smartAuthorizationHelper.ts` and `smartScopeHelper.ts`.

`smartScopeHelper` contains all the methods related to parsing and verifying scopes.
`smartAuthorizationHelper` contains the method `authorizeResource` and the method `authorizeResource` depends on `isLocalUserInJsonAsReference`. This file also contains the generic method `getFhirUser`. 

Methods that are required by `smartHandler` are exported, while the other helper methods are private.


src/smartScopeHelper.ts
```diff
@@ -0,0 +1,157 @@
...
+    scopeRule: ScopeRule,
+    reqOperation: TypeOperation | SystemOperation,
+    reqResourceType?: string,
+): (TypeOperation | SystemOperation)[] {
+    let validOperations: (TypeOperation | SystemOperation)[] = [];
+    if (smartScope.scopeType === 'launch') {
+        const launchSmartScope = <LaunchSmartScope>smartScope;
+        // TODO: should launch have access to only certain resourceTypes?
+        validOperations = scopeRule.launch[launchSmartScope.launchType ?? 'launch'];
+    } else if ((<ClinicalSmartScope>smartScope).scopeType) {
```
rsmayda: Do we need this to be else if, could this just be an `else`? isn't `scopeType` a required field for `SmartScope`? (this is probably copy and pasted over, but still curious)

src/smartScopeHelper.ts
```diff
@@ -0,0 +1,157 @@
...
+    bulkDataAuth?: BulkDataAuth,
+): boolean {
+    for (let i = 0; i < scopes.length; i += 1) {
+        const scope = scopes[i];
+        let smartScope: SmartScope | undefined;
+        try {
+            smartScope = convertScopeToSmartScope(scope);
+        } catch (e) {
+            // Caused by trying to convert non-SmartScope to SmartScope, for example converting scope 'openid' or 'profile'
+            // We don't need to check non-SmartScope
```
rsmayda: If this is the case could we add a `continue` here to more strongly indicate we are skipping this scope

src/smartScopeHelper.ts
```diff
@@ -0,0 +1,157 @@
...
+    resourceType?: string,
+    bulkDataAuth?: BulkDataAuth,
+): boolean {
+    for (let i = 0; i < scopes.length; i += 1) {
+        const scope = scopes[i];
+        let smartScope: SmartScope | undefined;
+        try {
+            smartScope = convertScopeToSmartScope(scope);
+        } catch (e) {
+            // Caused by trying to convert non-SmartScope to SmartScope, for example converting scope 'openid' or 'profile'
```
rsmayda: minor clean up
```suggestion
            // Caused by trying to convert non-SmartScope to SmartScope, for example converting non-SMART scope 'openid'
```

src/smartScopeHelper.ts
```diff
@@ -0,0 +1,157 @@
...
+            // We don't need to check non-SmartScope
+        }
+        if (smartScope) {
```
rsmayda: Could remove this `if` if we add a continue to the `catch`

src/smartAuthorizationHelper.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    test('resourceType is Practitioner', () => {
+        expect(authorizeResource(fhirUser, {}, 'https://fhirServer.com/')).toEqual(true);
+    });
+    describe('fhirUser resourceType matches resourceType of resource', () => {
+        test('fhirUser id matches resource id', () => {
+            const patientFhirUser: FhirUser = {
+                hostname: 'https://fhirServer.com/',
+                id: '1234',
+                resourceType: 'Patient',
+            };
```
rsmayda: Could move the `const` to be at the describe level and share between tests

src/smartScopeHelper.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        write: [],
+    },
+    launch: {
+        launch: [],
+        patient: [],
+        encounter: [],
+    },
+};
+describe('areScopesSufficient', () => {
+    test('scope is sufficient to read Observation', () => {
```
rsmayda: Could you add test for `patient` scope as well?

src/smartScopeHelper.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        const clonedScopeRule = clone(scopeRule);
+        clonedScopeRule.user.read = ['read'];
+
+        expect(areScopesSufficient(['user/Observation.read'], 'read', clonedScopeRule, 'Observation')).toEqual(true);
+
+        expect(areScopesSufficient(['user/*.read'], 'read', clonedScopeRule, 'Observation')).toEqual(true);
+    });
+    test('scope is NOT sufficient to read Observation', () => {
+        const clonedScopeRule = clone(scopeRule);
+        clonedScopeRule.user.read = ['read'];
```
rsmayda: Could we add a test where the scope rules does not have sufficient operations i.e.

`clonedScopeRule.user.read = ['search-type'];`

src/smartScopeHelper.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    });
+
+    test('scope is sufficient for bulk data access', () => {
+        const clonedScopeRule = clone(scopeRule);
+        clonedScopeRule.user.read = ['read'];
+        const bulkDataAuth: BulkDataAuth = { operation: 'initiate-export', exportType: 'system' };
+        expect(areScopesSufficient(['user/*.read'], 'read', clonedScopeRule, undefined, bulkDataAuth)).toEqual(true);
+    });
+
+    test('scope is NOT sufficient for bulk data access', () => {
```
rsmayda: Lets add a reason as to why it is not sufficient in this case (there are many not sufficient possibilities)

src/smartScopeHelper.test.ts
rsmayda: Also lets add a reason for failure if there is more than 1 failure case

rsmayda (commented):
Thanks for the refactor mostly minor/test related comments

src/smartScopeHelper.ts
nguyen102: Yep, you're right. This could be changed to an `else`.

src/smartScopeHelper.ts
nguyen102: Adding a `continue` gives us this `eslint` [error](https://eslint.org/docs/rules/no-continue).

I've added the `if (smartScope)` code block into the try statement though. That'll make it clearer that we're skipping non-SmartScope.

src/smartScopeHelper.ts
nguyen102: updated

src/smartScopeHelper.ts
nguyen102: I moved this if into the try statement. Using `continue` results in an `eslint` error.

src/smartAuthorizationHelper.test.ts
nguyen102: updated

src/smartScopeHelper.test.ts
nguyen102: added

src/smartScopeHelper.test.ts
nguyen102: Added the additional test.

Unfortunately jest doesn't natively allow passing in custom error messages. [(more details here)](https://stackoverflow.com/questions/45348083/how-to-add-custom-message-to-jest-expect). Therefore I'll just split the failure case into 2 tests instead of 1.

src/smartScopeHelper.test.ts
nguyen102: added

rsmayda (approved):

PR #13, opened 2021-01-09
feat: implement patient scope; remove system & launch scopes

rsmayda:
### smartAuthorizationHelper
- Change `FhirUser` -> `FhirResource` in the future there will be other launch contexts that we will need to support and they aren't really a `user` e.g. `launch/encounter`
### smartConfig
- Remove `launch` & `system` scopes
- Remove `scopeValueType` config option in favor of programmatically determining if the scopes are in array or string format
### smartHandler
- Remove `typesWithWriteAccess` - this was a forced customization if customer do not want to support writes they can specify in the `scopeRules`
- verifyAccessToken: it is now expected to have `fhirUserClaim` only if request has the `user/` scope and to have `launch_response_patient` only if request has `patient/` scope. If the scope is there but is missing the claim we ignore the said scope, since we cannot verify it.
- verifyAccessToken:  `userIdentity` will now also contain `fhirUserObject` & `patientLaunchContext` which are objects of the original `fhirUser` & `launch_patient` claim
- getSearchFilterBasedOnIdentity: add a filter for both `fhirUserObject` & `patientLaunchContext` if present
- isBundleRequestAuthorized: only use scopes that can be used (i.e. filter out user/ or patient/ if missing context)
### smartScopeHelper
- Remove system scope and launch scope
- switch `areScopesSufficient` -> `isScopeSufficient` which gives flexibility to use .filter
### TESTS
- smartHandler.test: was updated to not use a raw access_token and jwt's decode method making it much easier to understand and create new test-cases (this resulted in a lot of changes though)

#### Coverage information
after running `yarn run test-coverage` below are the results:
```sh
All files                    |   99.46 |     95.8 |     100 |   99.44 |
 smartAuthorizationHelper.ts |     100 |      100 |     100 |     100 |
 smartHandler.ts             |   99.12 |    94.05 |     100 |   99.08 | 106
 smartScopeHelper.ts         |     100 |    97.67 |     100 |     100 | 58

Test Suites: 4 passed, 4 total
Tests:       178 passed, 178 total
Snapshots:   0 total
Time:        14.759 s
Ran all test suites.
Done in 18.72s.
```

### DIFF
--- a/src/regExpressions.test.ts
+++ b/src/regExpressions.test.ts
@@ -8,34 +8,22 @@ import { FHIR_USER_REGEX, FHIR_RESOURCE_REGEX } from './smartAuthorizationHelper
 
 describe('CLINICAL_SCOPE_REGEX', () => {
     const testCases = [
-        ['patient', 'Patient', 'read', true],
-        ['user', 'Patient', 'read', true],
-        ['system', 'Patient', 'read', false],
-        ['patient', 'Patient', 'write', true],
-        ['patient', 'Patient', '*', true],
-        ['patient', 'Observation', 'read', true],
-        ['patient', 'FakeResource', 'write', true],
-        ['patient', '*', 'write', true],
-        ['patient', 'uncapitalizedResource', 'write', false],
-        ['fake', 'Patient', 'write', false],
-        ['patient', 'Patient', 'fake', false],
-        ['patient', 'Patient1', '*', false],
-        ['', 'Patient', 'read', false],
-        ['patient', '', 'read', false],
-        ['patient', 'Patient', '', false],
+        ['patient', 'Patient', 'read'],
+        ['user', 'Patient', 'read'],
+        ['patient', 'Patient', 'write'],
+        ['patient', 'Patient', '*'],
+        ['patient', 'Observation', 'read'],
+        ['patient', 'FakeResource', 'write'],
+        ['patient', '*', 'write'],
     ];
-    test.each(testCases)('CASE: %p/%p.%p; expect: %p', async (scopeType, scopeResourceType, accessType, isSuccess) => {
+    test.each(testCases)('CASE: %p/%p.%p; expect: %p', async (scopeType, scopeResourceType, accessType) => {
         const expectedStr = `${scopeType}/${scopeResourceType}.${accessType}`;
         const actualMatch = expectedStr.match(CLINICAL_SCOPE_REGEX);
-        if (isSuccess) {
-            expect(actualMatch).toBeTruthy();
-            expect(actualMatch!.groups).toBeTruthy();
-            expect(actualMatch!.groups!.scopeType).toEqual(scopeType);
-            expect(actualMatch!.groups!.scopeResourceType).toEqual(scopeResourceType);
-            expect(actualMatch!.groups!.accessType).toEqual(accessType);
-        } else {
-            expect(actualMatch).toBeFalsy();
-        }
+        expect(actualMatch).toBeTruthy();
+        expect(actualMatch!.groups).toBeTruthy();
+        expect(actualMatch!.groups!.scopeType).toEqual(scopeType);
+        expect(actualMatch!.groups!.scopeResourceType).toEqual(scopeResourceType);
+        expect(actualMatch!.groups!.accessType).toEqual(accessType);
     });
     const uniqueTestCases = [
         ['patient.Patient/read'],
@@ -43,6 +31,14 @@ describe('CLINICAL_SCOPE_REGEX', () => {
         ['patient/Patient.read patient/Patient.read'],
         ['launch/patient'],
         ['patient.Patient/read '],
+        ['patient/uncapitalizedResource.write'],
+        ['fake/Patient.write'],
+        ['patient/Patient.fake'],
+        ['patient/Patient1.*'],
+        ['/Patient.read'],
+        ['patient/.read'],
+        ['patient/Patient.'],
+        ['system/Patient.read'],
     ];
     test.each(uniqueTestCases)('CASE: %p; expect: false', async scope => {
         const actualMatch = scope.match(CLINICAL_SCOPE_REGEX);
@@ -52,40 +48,25 @@ describe('CLINICAL_SCOPE_REGEX', () => {
 
 describe('FHIR_USER_REGEX', () => {
     const testCases = [
-        ['https://fhir.server.com/dev/', 'Patient', 'id', true],
-        ['http://fhir.server.com/dev/', 'Patient', 'id', true],
-        ['http://fhir.server.com/dev-.:/%/$/2/', 'Patient', 'id', true],
-        ['http://localhost/projectname/', 'Patient', 'id', true],
-        ['http://127.0.0.1/project_name/', 'Patient', 'id', true],
-        ['https://127.0.0.1:8080/project_name/', 'Patient', 'id', true],
-        ['https://fhir.server.com/dev/', 'Practitioner', 'id', true],
-        ['https://fhir.server.com/dev/', 'RelatedPerson', 'id', true],
-        ['https://fhir.server.com/dev/', 'Person', 'id', true],
-        ['https://fhir.server.com/dev/', 'Patient', 'idID1234-123.aBc', true],
-        ['', 'Patient', 'id', false],
-        ['https://fhir.server.com/dev/', '', 'id', false],
-        ['https://fhir.server.com/dev/', 'Patient', '', false],
-        ['fhir.server.com/dev/', 'Patient', 'id', false],
-        ['https://fhir.server.com/dev', 'Patient', 'id', false],
-        ['127.0.0.1/project_name', 'Patient', 'id', false],
-        ['https://fhir.server.com/dev/', 'Observation', 'id', false],
-        ['https://fhir.server.com/dev/', 'Patient', 'i_d', false],
-        ['https://fhir.server.com/dev/', 'Patient', 'i#d', false],
-        ['https://fhir.server.com/dev/', 'Patient', 'id ', false],
-        [' https://fhir.server.com/dev/', 'Patient', 'id', false],
+        ['https://fhir.server.com/dev/', 'Patient', 'id'],
+        ['http://fhir.server.com/dev/', 'Patient', 'id'],
+        ['http://fhir.server.com/dev-.:/%/$/2/', 'Patient', 'id'],
+        ['http://localhost/projectname/', 'Patient', 'id'],
+        ['http://127.0.0.1/project_name/', 'Patient', 'id'],
+        ['https://127.0.0.1:8080/project_name/', 'Patient', 'id'],
+        ['https://fhir.server.com/dev/', 'Practitioner', 'id'],
+        ['https://fhir.server.com/dev/', 'RelatedPerson', 'id'],
+        ['https://fhir.server.com/dev/', 'Person', 'id'],
+        ['https://fhir.server.com/dev/', 'Patient', 'idID1234-123.aBc'],
     ];
-    test.each(testCases)('CASE: %p%p/%p; expect: %p', async (hostname, resourceType, id, isSuccess) => {
+    test.each(testCases)('CASE: %p%p/%p; expect: %p', async (hostname, resourceType, id) => {
         const expectedStr = `${hostname}${resourceType}/${id}`;
         const actualMatch = expectedStr.match(FHIR_USER_REGEX);
-        if (isSuccess) {
-            expect(actualMatch).toBeTruthy();
-            expect(actualMatch!.groups).toBeTruthy();
-            expect(actualMatch!.groups!.hostname).toEqual(hostname);
-            expect(actualMatch!.groups!.resourceType).toEqual(resourceType);
-            expect(actualMatch!.groups!.id).toEqual(id);
-        } else {
-            expect(actualMatch).toBeFalsy();
-        }
+        expect(actualMatch).toBeTruthy();
+        expect(actualMatch!.groups).toBeTruthy();
+        expect(actualMatch!.groups!.hostname).toEqual(hostname);
+        expect(actualMatch!.groups!.resourceType).toEqual(resourceType);
+        expect(actualMatch!.groups!.id).toEqual(id);
     });
     const uniqueTestCases = [
         ['patient/Patient.read'],
@@ -93,6 +74,17 @@ describe('FHIR_USER_REGEX', () => {
         ['just-an-id-1234'],
         ['Patient'],
         ['https://fhir.server.com/dev/'],
+        ['Patient/id'],
+        ['https://fhir.server.com/dev//id'],
+        ['https://fhir.server.com/dev/Patient/'],
+        ['fhir.server.com/dev/Patient/id'],
+        ['https://fhir.server.com/devPatient/id'],
+        ['127.0.0.1/project_namePatient/id'],
+        ['https://fhir.server.com/dev/Observation/id'],
+        ['https://fhir.server.com/dev/Patient/i_d'],
+        ['https://fhir.server.com/dev/Patient/i#d'],
+        ['https://fhir.server.com/dev/Patient/id '],
+        [' https://fhir.server.com/dev/Patient/id'],
     ];
     test.each(uniqueTestCases)('CASE: %p; expect: false', async scope => {
         const actualMatch = scope.match(FHIR_USER_REGEX);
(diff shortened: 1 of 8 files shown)


src/regExpressions.test.ts
```diff
@@ -109,18 +75,57 @@ describe('FHIR_USER_REGEX', () => {
         [' https://fhir.server.com/dev/', 'Patient', 'id', false],
     ];
     test.each(testCases)('CASE: %p%p/%p; expect: %p', async (hostname, resourceType, id, isSuccess) => {
-        for (let i = 0; i < testCases.length; i += 1) {
```
rsmayda, line 112: removing this needless `for`

src/smartHandler.ts
```diff
@@ -180,27 +211,22 @@ export class SMARTHandler implements Authorization {
         for (let i = 0; i < request.userIdentity.scopes.length; i += 1) {
             const scope = request.userIdentity.scopes[i];
             try {
-                const smartScope = convertScopeToSmartScope(scope);
                 // We only get allowedResourceTypes for ClinicalSmartScope
-                if (['patient', 'user', 'system'].includes(smartScope.scopeType)) {
```
rsmayda, line 185: removed this `if`

src/smartHandler.ts
```diff
@@ -180,27 +211,22 @@ export class SMARTHandler implements Authorization {
...
-                    const clinicalSmartScope = <ClinicalSmartScope>smartScope;
-                    const validOperations = getValidOperationsForScopeTypeAndAccessType(
-                        clinicalSmartScope.scopeType,
-                        clinicalSmartScope.accessType,
-                        this.config.scopeRule,
-                    );
-                    if (validOperations.includes(request.operation)) {
-                        const allowedResourcesForScope: string[] =
-                            this.fhirVersion === '4.0.1' ? BASE_R4_RESOURCES : BASE_STU3_RESOURCES;
-                        if (['patient', 'user', 'system'].includes(clinicalSmartScope.scopeType)) {
```
rsmayda, line 195: removed this `if`

src/smartScopeHelper.ts
```diff
@@ -118,33 +94,53 @@ function isSmartScopeSufficientForBulkDataAccess(
...
     return false;
 }
+
+/**
+ * Remove scopes that do not have the required information to be useful or unused scopes. For example:
+ * - Without the `fhirUser` claim the 'user' scopes cannot be validated
+ * - Without the `launch_response_patient` claim the 'patient' scopes cannot be validated
+ * - Scopes like profile, launch or fhirUser will be filtered out as well
+ */
+export function filterScopes(
```
nguyen102: nitpick: Can we rename this to something like `filterOutUnusableScope`?

nguyen102 (dismissed):
Great job! This looks good! One small commet about renaming filterScopes

src/smartHandler.test.ts
```diff
@@ -252,319 +246,322 @@ describe('constructor', () => {
...
+        ],
     ];
 
     const authZHandler: SMARTHandler = new SMARTHandler(authZConfig, apiUrl, '4.0.1');
-    test.each(arrayScopesCases)('CASE: %p', (_firstArg, request, isValid) => {
-        mock.onGet(authZConfig.userInfoEndpoint).reply(200, patientIdentityWithoutScopes);
+    test.each(cases)('CASE: %p', (_firstArg, request, decodedAccessToken, isValid) => {
+        mock.onGet(authZConfig.userInfoEndpoint).reply(200, decodedAccessToken);
+        const { decode } = jwt as jest.Mocked<typeof import('jsonwebtoken')>;
+        decode.mockReturnValue(<{ [key: string]: any }>decodedAccessToken);
```
carvantes, line 408: Why are we mocking this? I think it's better to just encode `decodedAccessToken` into a JWT with `jwt.sign`

https://www.npmjs.com/package/jsonwebtoken#jwtsignpayload-secretorprivatekey-options-callback

src/regExpressions.test.ts
```diff
@@ -109,18 +75,57 @@ describe('FHIR_USER_REGEX', () => {
...
+    const testCases = [
+        ['https://fhir.server.com/dev/', 'Patient', 'id', true, true],
+        ['http://fhir.server.com/dev-.:/%/$/2/', 'Observation', 'id', true, true],
+        ['http://localhost/projectname/', 'Encounter', 'id', true, true],
+        ['https://127.0.0.1:8080/project_name/', 'Patient', 'id', true, true],
+        ['https://fhir.server.com/dev/', 'Patient', 'idID1234-123.aBc', true, true],
+        ['', 'Patient', 'id', true, false],
+        ['', 'Encounter', 'id', true, false],
+        ['fhir.server.com/dev/', 'Patient', 'id', true, false],
+        ['127.0.0.1/project_name', 'Patient', 'id', true, false],
```
carvantes: IMO using tests arrays to branch test assertions is a bad practice, it is hard to understand what all those `true` and `false` represent. 

I think it's better to keep the array focused on test data only, with the assumption that it's just different test data that goes through the same test code.

here `hasHostname` can be derived from `hostname`, and it'd be better to have 2 test suites, on for success and one for failure instead of relying on `isSuccess` (also, all tests have `isSuccess=true`, so `isSuccess` is useless)

src/smartScopeHelper.test.ts
```diff
@@ -79,17 +71,107 @@ describe.each(areScopesSufficientCases)('%s: areScopesSufficient', (scopeType: s
 
         const bulkDataAuth: BulkDataAuth = { operation: 'initiate-export', exportType: 'system' };
         expect(
-            areScopesSufficient([`${scopeType}/Observation.read`], 'read', clonedScopeRule, undefined, bulkDataAuth),
+            isScopeSufficient(`${scopeType}/Observation.read`, clonedScopeRule, 'read', undefined, bulkDataAuth),
         ).toEqual(false);
     });
+    test('scope is sufficient to do a search-system', () => {
+        const clonedScopeRule = clone(scopeRule);
+        clonedScopeRule[scopeType].read = ['search-system'];
```
carvantes: I think it's best to avoid using `clone` here. I think it was a bad idea to add that function to the interface package and we should eventually remove it.

A quick fix would be to use a function that generates the empty scope rule instead:
```js
const emptyScopeRule = () => ({
    patient: {
        read: [],
        write: [],
    },
    user: {
        read: [],
        write: [],
    },
});
```

src/smartScopeHelper.test.ts
```diff
@@ -79,17 +71,107 @@ describe.each(areScopesSufficientCases)('%s: areScopesSufficient', (scopeType: s
...
+            'user/*.read',
+            'fake',
+        ]);
+    });
+});
+describe('filterScopes', () => {
+    test('no filter occuring', () => {
+        const clonedScopeRule = clone(scopeRule);
+        clonedScopeRule.user.read = ['read'];
+        clonedScopeRule.patient.read = ['read'];
```
carvantes: It's probably easier and clearer to just do:
```js
const scopeRule = {
    patient: {
        read: ['read'],
        write: [],
    },
    user: {
        read: ['read'],
        write: [],
    },
}
```

btw, I just realized that the scope rule is weird in that you can have `patient.write['read']` and `patient.read['write']`. What does it mean to have "write" permissions on the "read" operation?

src/smartConfig.ts
```diff
@@ -15,65 +14,43 @@ export interface ClinicalSmartScope {
...
-};
 
-// Determines what each scope has access to
+/**
+ * Determines what each scope has access to
+ *
+ *  @example
+ * {
+ *      patient: {
+ *          read: allReadOperations,
```
carvantes: what does `allReadOperations` means? I think that the `@example` should be self-contained

src/smartAuthorizationHelper.test.ts
```diff
@@ -1,23 +1,55 @@
...
+            id: '1234',
+            resourceType: 'Practitioner',
+        });
+        expect(getFhirResource('https://fhirServer1234.com/Organization/1234', defaultHostname)).toEqual({
+            hostname: 'https://fhirServer1234.com/',
+            id: '1234',
+            resourceType: 'Organization',
+        });
+    });
+    test('valid fhirResource;; With default hostname', () => {
```
carvantes: typo: `;;`

src/smartAuthorizationHelper.ts
```diff
@@ -1,42 +1,54 @@
...
     throw new UnauthorizedError("Requester's identity is in the incorrect format");
 }
+export function getFhirResource(resourceValue: string, defaultHostname: string): FhirResource {
+    const match = resourceValue.match(FHIR_RESOURCE_REGEX);
+    if (match) {
+        // eslint-disable-next-line prefer-const
+        let { hostname, resourceType, id } = match.groups!;
+        if (!hostname) {
+            hostname = defaultHostname;
+        }
```
carvantes: I think that we should avoid disabling eslint rules unless we have a strong reason to do so. 
```suggestion
        const { resourceType, id } = match.groups!;
        const hostname = match.groups!.hostname ?? defaultHostname;
```

src/smartAuthorizationHelper.ts
```diff
@@ -1,42 +1,54 @@
-import { KeyValueMap, UnauthorizedError } from 'fhir-works-on-aws-interface';
-import { IdentityType } from './smartConfig';
+import { UnauthorizedError } from 'fhir-works-on-aws-interface';
 
 export const FHIR_USER_REGEX = /^(?<hostname>(http|https):\/\/([A-Za-z0-9\-\\.:%$_]*\/)+)(?<resourceType>Person|Practitioner|RelatedPerson|Patient)\/(?<id>[A-Za-z0-9\-.]+)$/;
+export const FHIR_RESOURCE_REGEX = /(?<hostname>^(http|https):\/\/([A-Za-z0-9\-\\.:%$_]*\/)+)?(?<resourceType>[A-Z][a-zA-Z]+)\/(?<id>[A-Za-z0-9\-.]+)$/;
```
carvantes: looks like the `^` anchor is misplaced. This regex matches `SomeRandomPrefixPatient/id`

carvantes (changes requested):

src/smartAuthorizationHelper.ts
```diff
@@ -1,42 +1,54 @@
...
     return (
-        jsonStr.includes(`"reference":"${fhirUser.hostname}${fhirUser.resourceType}/${fhirUser.id}"`) ||
-        jsonStr.includes(`"reference":"${fhirUser.resourceType}/${fhirUser.id}"`)
+        jsonStr.includes(`"reference":"${fhirResource.hostname}${fhirResource.resourceType}/${fhirResource.id}"`) ||
+        jsonStr.includes(`"reference":"${fhirResource.resourceType}/${fhirResource.id}"`)
     );
 }
 
-export function authorizeResource(fhirUser: FhirUser, resource: any, apiUrl: string): boolean {
+export function authorizeResource(fhirResource: FhirResource, resource: any, apiUrl: string): boolean {
```
carvantes: nit: rename to something like `hasReferenceToResource`. This function doesn't really "authorizes" something

carvantes: Actually this rename is not trivial since this function has the weird condition that does not deal with references:
```
if (resourceType === 'Practitioner') {
        return true;
    }
```

Still, I believe that the above condition can be removed since it is already checked outside this function as part of the `this.adminAccessTypes.includes` check

src/smartHandler.test.ts
rsmayda: mocking it as it makes test generation easier, previously we just used a raw opaque JWT token, now we can use objects

Using jwt.sign is an option but it also ties this unit test to jwt (which isn't necessarily bad). I don't have a strong preference but would prefer to leave it as is as @nguyen102 is moving the code from using decode -> verify and that has probably a better use-case for using this jwt.sign approach

src/smartScopeHelper.test.ts
rsmayda: > What does it mean to have "write" permissions on the "read" operation?

Yep that is right. I changed the type so that it will prevent these sort of mistakes

src/smartHandler.test.ts
nguyen102: Given this use case, I don't see any issue with mocking. 

It's simpler than using signing. To use `jwt.sign` correctly, we would need to sign using the `RS256` algorithm instead of the default `HS256` algorithm. The `RS256` algorithm requires a private key, which we need to store and import into this test.

If mocking accomplish the same objective, which is to make the payload of the JWT easy to read, we should just use mocking.

src/regExpressions.test.ts
rsmayda: cleaned up in this file

src/smartScopeHelper.test.ts
rsmayda: updated!

src/smartConfig.ts
rsmayda: fixed

src/smartAuthorizationHelper.ts
rsmayda: fixed

src/smartAuthorizationHelper.ts
rsmayda: For FHIR_RESOURCE_REGEX  we are okay with using a default hostname (if one is not provided), but if one is provided it should be legal ie:
- Good:
  - Patient/id
  - http://new.api.com/Patient/id
- Bad:
  - Fake things/Patient/id

I did have to move `^` but do note the `?` making hostname part optional

src/smartAuthorizationHelper.ts
rsmayda: fixed

src/smartAuthorizationHelper.test.ts
rsmayda: fixed

src/regExpressions.test.ts
```diff
@@ -8,41 +8,37 @@ import { FHIR_USER_REGEX, FHIR_RESOURCE_REGEX } from './smartAuthorizationHelper
...
+        ['patient', 'Patient', 'read'],
+        ['user', 'Patient', 'read'],
+        ['patient', 'Patient', 'write'],
+        ['patient', 'Patient', '*'],
+        ['patient', 'Observation', 'read'],
+        ['patient', 'FakeResource', 'write'],
+        ['patient', '*', 'write'],
     ];
-    test.each(testCases)('CASE: %p/%p.%p; expect: %p', async (scopeType, scopeResourceType, accessType, isSuccess) => {
+    test.each(testCases)('CASE: %p/%p.%p; expect: %p', async (scopeType, scopeResourceType, accessType) => {
```
carvantes: You need to update the test names as well. It still has 4 instances of `%p`

src/regExpressions.test.ts
```diff
@@ -52,47 +48,43 @@ describe('CLINICAL_SCOPE_REGEX', () => {
...
+        ['http://localhost/projectname/', 'Patient', 'id'],
+        ['http://127.0.0.1/project_name/', 'Patient', 'id'],
+        ['https://127.0.0.1:8080/project_name/', 'Patient', 'id'],
+        ['https://fhir.server.com/dev/', 'Practitioner', 'id'],
+        ['https://fhir.server.com/dev/', 'RelatedPerson', 'id'],
+        ['https://fhir.server.com/dev/', 'Person', 'id'],
+        ['https://fhir.server.com/dev/', 'Patient', 'idID1234-123.aBc'],
     ];
-    test.each(testCases)('CASE: %p%p/%p; expect: %p', async (hostname, resourceType, id, isSuccess) => {
+    test.each(testCases)('CASE: %p%p/%p; expect: %p', async (hostname, resourceType, id) => {
```
carvantes: update test names as well

src/smartHandler.test.ts
carvantes: yeah, the timing of these changes make it trickier to coordinate. I agree that using real JWT tokens here can be cumbersome with the switch to asymmetric keys. I leave it up to you 2 to decide.

IMO mocking is ok, but mocking `jsonwebtoken.decode` is too low level. It'd be better to mock an higher level abstraction like `smartAuthorizationHelper.verifyJwtToken` but that's on Tim's new PR so it's probably ok to leave as is.

src/smartAuthorizationHelper.ts
```diff
@@ -1,42 +1,51 @@
...
+    const { hostname, resourceType, id } = fhirResource;
+    if (hostname !== apiUrl) {
         // If requester is not from this FHIR Server they must be a fully qualified reference
-        return jsonStr.includes(`"reference":"${fhirUser.hostname}${fhirUser.resourceType}/${fhirUser.id}"`);
+        return jsonStr.includes(`"reference":"${hostname}${resourceType}/${id}"`);
     }
-    if (fhirUser.resourceType === 'Practitioner') {
+    if (resourceType === 'Practitioner') {
         return true;
     }
```
carvantes, line 45: should this be updated to use the `SMARTHandler.adminAccessTypes` instead of hardcoding `'Practitioner'`?

I'm actually not 100% sure about what `adminAccessTypes` does

carvantes (approved):
looks good. Just one more comment regarding `adminAccessTypes`. It's about existing code so I'm not blocking on it.

PR #22, opened 2021-02-06
feat: make reference checking more restrictive and abid by FHIR

rsmayda:
Description of changes:
- A 'valid' reference now is not only a valid reference path but also a valid reference resourceType
  - i.e. For a Patient the `generalPractitioner` reference can only be a Practitioner or an Organization
- Added script to create reference matrix json. Rough reference mapping: 
```json
{ "sourceResourceType":{ "requestorResourceType": ["paths"] }
//Example
{"Account":{"Patient":["subject","guarantor.party"]}}
```

Test coverage
```bash
All files                    |   99.57 |    96.43 |     100 |   99.55 |
 smartAuthorizationHelper.ts |     100 |      100 |     100 |     100 |
 smartHandler.ts             |   99.12 |    94.32 |     100 |   99.07 | 97
 smartScopeHelper.ts         |     100 |    97.67 |     100 |     100 | 62

Test Suites: 4 passed, 4 total
Tests:       192 passed, 192 total
Snapshots:   0 total
Time:        16.851 s
Ran all test suites.
```


src/smartHandler.ts
```diff
@@ -1,288 +1,294 @@
...
-    }
-
-    async authorizeAndFilterReadResponse(request: ReadResponseAuthorizedRequest): Promise<any> {
-        const { fhirUserObject, patientLaunchContext } = request.userIdentity;
-        const { operation, readResponse } = request;
-        // If request is a search treat the readResponse as a bundle
-        if (SEARCH_OPERATIONS.includes(operation)) {
-            const entries = (readResponse.entry ?? []).filter(
-                (entry: { resource: any }) =>
-                    (fhirUserObject && hasReferenceToResource(fhirUserObject, entry.resource, this.apiUrl)) ||
```
rsmayda, line 255: Please enable ignore white space changes -- this file was converted from CRLF to LF line endings

src/smartAuthorizationHelper.ts
```diff
@@ -32,29 +35,90 @@ export function getFhirResource(resourceValue: string, defaultHostname: string):
     throw new UnauthorizedError('Resource is in the incorrect format');
 }
 
-function isLocalResourceInJsonAsReference(jsonStr: string, fhirResource: FhirResource): boolean {
-    return (
-        jsonStr.includes(`"reference":"${fhirResource.hostname}/${fhirResource.resourceType}/${fhirResource.id}"`) ||
-        jsonStr.includes(`"reference":"${fhirResource.resourceType}/${fhirResource.id}"`)
-    );
+function isReqestorReferenced(
```
carvantes: typo

scripts/generateResourceReferenceMatrixFile.ts
```diff
@@ -0,0 +1,115 @@
...
+const compile = (resources: any[], fhirVersion: string) => {
+    const filter = resources.filter(
+        resource => resource.baseDefinition === 'http://hl7.org/fhir/StructureDefinition/DomainResource',
+    );
+
+    const result: Result = {};
+    if (fhirVersion.startsWith('4')) {
+        filter.forEach(resource => {
+            return resource.snapshot.element
+                .filter((element: Element) => element.type && element.type[0].code === 'Reference')
```
carvantes: I think that this is not handling properly fields with multiple types. only the 1st type is checked on `element.type[0].code`

There are some fields that among there multiple types have a the reference type. i.e. `ExplanationOfBenefit.accident.location[x]` can be `ExplanationOfBenefit.accident.locationReference`

Not sure if any of those fields are meaningful for the SMART scopes we support though.

https://www.hl7.org/fhir/explanationofbenefit-definitions.html#ExplanationOfBenefit.accident.location_x_

scripts/generateResourceReferenceMatrixFile.ts
```diff
@@ -0,0 +1,115 @@
...
+                .forEach((element: Element) =>
+                    element.type[0].targetProfile.forEach((target: string) => {
+                        const sourceType = resource.type;
+                        const requestorType = target.replace('http://hl7.org/fhir/StructureDefinition/', '');
+                        const path = element.id.replace(`${resource.type}.`, '').replace('[x]', '');
```
carvantes: dropping the `[x]` is also related to not handling properly fields with multiple types

scripts/generateResourceReferenceMatrixFile.ts
rsmayda: +1 yep your right I left some out fixing

scripts/generateResourceReferenceMatrixFile.ts
rsmayda: Thanks since we only care about refereneces here I will be replacing [x] with Reference

src/smartAuthorizationHelper.test.ts
```diff
@@ -71,40 +74,260 @@ describe('getFhirResource', () => {
...
+                organizationFhirUser,
+                { ...{ resourceType: 'Patient', id: '1' }, ...refObject },
+                apiUrl,
+                r4Version,
+            ),
+        ).toEqual(expectedValue);
+    });
+
+    const versions: FhirVersion[] = ['4.0.1', '3.0.1'];
+    describe.each(versions)('requestor is a Patient; Resources are single layer; FHIR Version %p', fhirVersion => {
```
carvantes, line 212: nit: what is the meaning of `Resources are single layer`?

carvantes (approved):

src/smartAuthorizationHelper.test.ts
rsmayda: references are not nested in arrays ie:
```json
{
link: { other: { reference: `Patient/${id}` } } 
}
```
vs nested
```json
{
link: [{ other: [{ reference: `Patient/${id}` }] } ]
}
```

PR #41, opened 2021-06-24
feat: Add support for `system` scope

rsmayda:
Description of changes:
- Treat `system` scopes as user scopes as describe:
> System scopes have the format system/(:resourceType|*).(read|write|*)– which conveys the same access scope as the matching user format user/(:resourceType|*).(read|write|*) https://hl7.org/fhir/uv/bulkdata/authorization/index.html#scopes
- Add testing to support new functionality
- In situations with `sub` is not provided in the JWT for system requests; an exception will be thrown

### DIFF
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -121,7 +121,6 @@ export class SMARTHandler implements Authorization {
         if (request.bulkDataAuth) {
             if (!userIdentity.sub) {
                 logger.error('A JWT token is without a `sub` claim; we cannot process the bulk action without one.');
-                // if there is a system scope and there is no sub set it to system
                 throw new UnauthorizedError('User does not have permission for requested operation');
             }
             if (
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -43,6 +43,8 @@ const logger = getComponentLogger();
 
 // eslint-disable-next-line import/prefer-default-export
 export class SMARTHandler implements Authorization {
+    private static defaultSystemId = '__system__';
+
     /**
      * If a fhirUser is of these resourceTypes they will be able to READ & WRITE without having to meet the reference criteria
      */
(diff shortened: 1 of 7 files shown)


src/smartScopeHelper.ts
```diff
@@ -13,7 +13,7 @@ export const SEARCH_OPERATIONS: (TypeOperation | SystemOperation)[] = [
     'history-system',
 ];
 
-export const CLINICAL_SCOPE_REGEX = /^(?<scopeType>patient|user)\/(?<scopeResourceType>[A-Z][a-zA-Z]+|\*)\.(?<accessType>read|write|\*)$/;
+export const CLINICAL_SCOPE_REGEX = /^(?<scopeType>patient|user|system)\/(?<scopeResourceType>[A-Z][a-zA-Z]+|\*)\.(?<accessType>read|write|\*)$/;
```
carvantes: nit: I *think* that scopes with `system` are not called "clinical scopes", so renaming this variable may make things clearer.

src/smartScopeHelper.test.ts
```diff
@@ -204,6 +214,25 @@ describe('filterOutUnusableScope', () => {
...
+            filterOutUnusableScope(
+                expectedScopes,
+                clonedScopeRule,
+                'read',
+                'Patient',
+                undefined,
+                undefined,
+                'fhirUser',
+            ),
+        ).toEqual([expectedScopes[0], expectedScopes[2]]);
```
carvantes: nit: It is very weird to have a var named `expectedScopes` be the input of the test and also not actually be the expected result.

Maybe try:
```js
const scopes = ['user/Patient.read', 'system/Obersvation.*', 'system/*.read'];
...

).toEqual(['user/Patient.read', 'system/*.read']);
```

src/smartHandler.ts
```diff
@@ -109,25 +111,32 @@ export class SMARTHandler implements Authorization {
...
-            }
-            const fhirUser = getFhirUser(fhirUserClaim);
-            if (fhirUser.hostname !== this.apiUrl || !this.bulkDataAccessTypes.includes(fhirUser.resourceType)) {
-                throw new UnauthorizedError('User does not have permission for requested operation');
+            if (
+                !usableScopes.some((scope: string) => {
+                    return scope.startsWith('system');
+                })
+            ) {
+                // if requrestor is relying on the "user" scope we need to verify they are coming from the correct endpoint & resourceType
```
carvantes: typo in `requrestor`

src/smartHandler.ts
```diff
@@ -109,25 +111,32 @@ export class SMARTHandler implements Authorization {
...
+                const fhirUser = getFhirUser(fhirUserClaim);
+                if (fhirUser.hostname !== this.apiUrl || !this.bulkDataAccessTypes.includes(fhirUser.resourceType)) {
+                    throw new UnauthorizedError('User does not have permission for requested operation');
+                }
+            } else if (!userIdentity.sub) {
+                // if there is a system scope and there is no sub set it to system
+                userIdentity.sub = SMARTHandler.defaultSystemId;
```
carvantes: I don't think this is safe.

the `sub` value is used to identify the owner of bulk export jobs: https://github.com/awslabs/fhir-works-on-aws-routing/blob/d00a9a01d7578eba24246b4eb0f56e67a0bb55aa/src/router/routes/exportRouteHelper.ts#L27

This would let system-scope-clients get the bulk export results from any other system-scope-client(they just need to know the jobId) since they all share the same `sub` value.

btw, I'm also not familiar with the use case where `sub` is missing. I thought that the `sub` claim was always present.

rsmayda: Yeah I struggled with this one -- `sub` is technically not required from OAuth2 -- perhaps the safest thing to do here is throw an error instead. ie. FWoA requires `sub` to be present in all access_tokens in bulk_access use case

carvantes: I agree, rejecting tokens without `sub` seems to be the way to go. i'm not sure what happens on bulk export if `sub` is undefined. Maybe it just crashes at some point.

carvantes (approved):

PR #43, opened 2021-07-08
feat: allow regex in SMARTConfig.expectedAudValue

carvantes:
Allow RegExp as expectedAudValue.

This enables use cases where the apiUrl is not static. The main use case is for multi-tenancy with tenant-specific urls, but there may be other use cases.

Only minor code changes are required since the `jsonwebtoken` package already accepts `string | RegExp` when verifying the aud value.
https://www.npmjs.com/package/jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback

**Note:** This is intended to support multi-tenancy but it is a generic change so no feature branch is being created.

### DIFF
--- a/src/smartConfig.ts
+++ b/src/smartConfig.ts
@@ -74,6 +74,9 @@ export interface SMARTConfig {
     scopeRule: ScopeRule;
     /**
      * Per SMART spec this is the 'aud' key found in the access_token
+     *
+     * Using the string type is recommended. RegExp can be useful when the audience is not static, such as in multi-tenant setups.
+     * Caution must be taken to avoid overly permissive regular expressions (e.g. avoid using .*). Use regular expressions that are as specific as possible to avoid allowing requests from unexpected audiences.
      */
     expectedAudValue: string | RegExp;
     /**
--- a/src/smartConfig.ts
+++ b/src/smartConfig.ts
@@ -75,7 +75,7 @@ export interface SMARTConfig {
     /**
      * Per SMART spec this is the 'aud' key found in the access_token
      */
-    expectedAudValue: string;
+    expectedAudValue: string | RegExp;
     /**
      * Per SMART spec this is the 'iss' key found in the access_token
      */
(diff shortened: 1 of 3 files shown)


src/smartConfig.ts
```diff
@@ -75,7 +75,7 @@ export interface SMARTConfig {
     /**
      * Per SMART spec this is the 'aud' key found in the access_token
      */
-    expectedAudValue: string;
+    expectedAudValue: string | RegExp;
```
rsmayda, line 81: Would passing in a reggex instead of a string result in a giving access incorrectly?
ie. audRegex = `apigw.com/tenant/.*/`

This would return true (i.e. validated) for both apigw.com/tenant/1/ & apigw.com/tenant/2/

carvantes: I don't think the access is incorrect. The point of using a regex is to allow tokens with different `aud` claims to be valid.

Both `apigw.com/tenant/1` `apigw.com/tenant/2` are valid `aud` claims when multi-tenancy is enabled (they indicate tenantId=1 and tenantId=2 respectively)

Care must be taken to use a narrow regex (e.g. avoid `.*`) to avoid unintended matches. For the default SMART config we are using: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/9d902a3f6b3f0b0921c128c30ab03d5a1f1ea10d/src/config.ts#L55

rsmayda: What is preventing someone from tenant = 1 accessing tenant = 2 data? There JWT would have
```json
{
"aud": "apigw.com/tenant/1/"
}
```

And their request would be `apigw.com/tenant/2/Patient/1234`

carvantes: Verifying that the tenantId from the token matches the one in the request url is done on the routing package:

https://github.com/awslabs/fhir-works-on-aws-routing/blob/f329d49162967db959924c7755b9cd3b0032025e/src/router/middlewares/setTenantId.ts#L48

rsmayda: Could we add a comment to the variable "using RegExp type is only advised when using multi-tenancy; using too loose of a RegExp can result in permitting requests from unexpected `audiences`"?

carvantes: sure, I just added the comment

rsmayda (approved):

PR #44, opened 2021-07-10
feat: allow token introspection as an authz option

rsmayda:
Description of changes:
- Added `IntrospectionOptions`
- Defaults to use introspection if configuration is given
- Made `jwksEndpoint` optional
- made `decodeJwtToken` its own method
- Changed the AuthZ errors from logging as `error` to log as `warn`

Coverage results:
```bash
All files                    |   98.93 |    95.85 |     100 |   98.91 |
 loggerBuilder.ts            |     100 |      100 |     100 |     100 |
 smartAuthorizationHelper.ts |   97.22 |    94.23 |     100 |    97.2 | 186,210,220
 smartHandler.ts             |     100 |    96.04 |     100 |     100 | 204,280,285,303
 smartScopeHelper.ts         |     100 |     97.5 |     100 |     100 | 62
```


src/smartHandler.ts
```diff
@@ -81,18 +82,32 @@ export class SMARTHandler implements Authorization {
...
+            );
+        } else if (this.jwksClient) {
+            decodedToken = await verifyJwtToken(
+                request.accessToken,
+                this.config.expectedAudValue,
+                this.config.expectedIssValue,
+                this.jwksClient,
+            );
+        } else {
+            throw Error('Authorization configuration not properly set up');
```
carvantes: Future devs will appreciate if this message is more specific. e.g.
`Authorization configuration not properly set up. Either 'tokenIntrospection' or 'jwksEndpoint' must be present`

src/smartConfig.ts
```diff
@@ -62,6 +62,22 @@ export interface UserIdentity extends KeyValueMap {
...
+    /**
+     * Your FHIR server's password, typically generated by your idp.
+     */
+    clientSecret: string;
+    /**
+     * The introspection url is determined by the incoming access_token's issuer field (`iss`) and this suffix. If the introspection endpoint is `cognito.com/oauth/v1/introspect` and the issuer is `cognito.com/`, you would want to set this suffix to be `oauth/v1/introspect`
+     * @example v1/introspect
+     */
+    introspectUrlSuffix: string;
+}
```
carvantes: Implementation is correct but this comment made me nervous at first.

I thought that fwoa could potentially make introspect requests to ANY url that happens to be the `iss` in a token, but then I realized that the token must first pass a check against the `expectedIssValue`.

I think it'd be good to clarify that introspection requests are ONLY made against the `expectedIssValue` domain.

**Actually, I think that the config would be more self-explanatory if we asked for the complete `introspectUrl` instead of just the suffix.
Right now the config seems weird to me since we allow exactly one `expectedIssValue` so we may as well provide a single `introspectUrl`.
The suffix could be convenient if we want to allow multiple `expectedIssValue` in the future but even then there's no guarantee that all endpoints would share the exact same suffix

carvantes (commented):
Overall looks good. Just a couple comments that may make the new config clearer

src/smartConfig.ts
rsmayda: I was on the fence about this too I think it is clearer using just `instropectUrl` we can iterate from here if customer need more options

carvantes (approved):

PR #45, opened 2021-07-15
feat: use fhirServiceBaseUrl from requests if present

carvantes:
Description of changes:

Use the `fhirServiceBaseUrl` from the incoming auth request instead of the static `this.apiUrl` set in the constructor.

This package uses the `apiUrl` to perform checks in launch_response/fhirUser claims and when building references to be used in search filters. In a multi-tenant setup the `apiUrl` can be different for each tenant. This change allows the client (i.e. the router package) to specify the api url to be used in the AuthZ check for a given request.

`fhirServiceBaseUrl` takes precedence over `this.apiUrl` as it is more specific(specific to the incoming request). This is a non-breaking change for single tenant setups.

See: 
- https://github.com/awslabs/fhir-works-on-aws-interface/pull/72
- https://github.com/awslabs/fhir-works-on-aws-routing/pull/100

### DIFF
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -65,6 +65,7 @@ export class SMARTHandler implements Authorization {
 
     /**
      * @param apiUrl URL of this FHIR service. Will be used to determine if a requestor is from this FHIR server or not
+     * when the request does not include a fhirServiceBaseUrl
      * @param adminAccessTypes a fhirUser from these resourceTypes they will be able to READ & WRITE without having to meet the reference criteria
      * @param bulkDataAccessTypes a fhirUser from these resourceTypes they will be able to do bulk data operations
      */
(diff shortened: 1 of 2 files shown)


src/smartHandler.ts
```diff
@@ -65,6 +65,7 @@ export class SMARTHandler implements Authorization {
 
     /**
      * @param apiUrl URL of this FHIR service. Will be used to determine if a requestor is from this FHIR server or not
+     * when the request does not include a fhirServiceBaseUrl
```
rsmayda, line 68: Does it make sense to keep the `apiUrl` parameter after this change? i.e. would it be better to force users to pass in `fhirServiceBaseUrl`?

Bingjiling (approved):
LGTM

src/smartHandler.ts
carvantes: Removing `apiUrl` would be a breaking change that I'd rather not do right now. 
The updated router does pass `fhirServiceBaseUrl` for all requests, but customers using `fhir-works-on-aws-authz-smart` without our router would be affected.

rsmayda (approved):

PR #48, opened 2021-08-02
fix: pass params in url for introspect jwt flow

medhost-jshaver:
Returning the decoded JWT token payload from introspectJwtToken function, and using string token parameter for call to introspect endpoint since it expects x-www-form-urlencoded content-type.

Issue #, if available:

Description of changes:
1. Changed introspectJwtToken method to return the decoded token payload since the introspect endpoint response doesn't necessarily return the JWT. 

For example, our introspect endpoint returns a payload with the "scope" key while the jwt uses "scp". 

2. change the token parameter type to a string for the actual introspect call. The Object structure did not seem to place nice with x-www-form-urlencoded content-type.

3. Updated test to reflect the introspectJwtToken returning the decodedToken payload and not the introspect response.

### DIFF
--- a/src/smartAuthorizationHelper.test.ts
+++ b/src/smartAuthorizationHelper.test.ts
@@ -534,7 +534,6 @@ describe('introspectJwtToken', () => {
             introspectJwtToken(jwt, expectedAudValue, expectedIssValue, introspectionOptions),
         ).resolves.toEqual({
             ...payload,
-            active: true,
         });
     });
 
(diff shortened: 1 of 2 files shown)


rsmayda (approved):
Code looks good!

@medhost-jshaver please [commit an empty commit](https://coderwall.com/p/vkdekq/git-commit-allow-empty) with a commit message following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary). Something like the below would work:
`fix: pass params in url for introspect jwt flow`

@carvantes or @Bingjiling to merge this in and release it.

PR #60, opened 2021-09-02
feat: update system export to only allow system scope by default

Bingjiling:
Issue #, if available:

Description of changes:

* Update system export to only allow system scope by default
* Add config value to allow user scope 
* Updated unit tests 


ssvegaraju (dismissed):
Looks good to me

src/smartScopeHelper.ts
```diff
@@ -86,8 +86,12 @@ function isSmartScopeSufficientForBulkDataAccess(
 ) {
     let bulkDataRequestHasCorrectScope = false;
     if (bulkDataAuth.exportType === 'system') {
+        const isScopeTypeSufficient =
+            process.env.ALLOW_USER_SCOPE_FOR_SYSTEM_EXPORT === 'true'
```
rsmayda: Could this be a constructor argument and not an env variable?

rsmayda (changes requested):

src/smartScopeHelper.ts
Bingjiling: Yes! That'll make it a lot simpler.

rsmayda (approved):

Issue #55, opened 2021-08-20
[Bug] Incorrect total value returned from authorizeAndFilterReadResponse


ghost:
**Describe the bug**
When the authorizeAndFilterReadResponse function runs it does a check to determine whether the user access by the hasAccessToResource function. If they do not it removes it rum the array. The return shows a total of 4 
"type": "searchset", "total": 4,
But in the entry section there are only 3 records.

**Expected behavior**
The total should match the number of entries being returned or potentially have another value that shows it.

rsmayda:
Thanks for reporting this. We will be adding this to our backlog!

ssvegaraju:
Hey! Just wanted to update on this: a fix for this has just been [merged into mainline](https://github.com/awslabs/fhir-works-on-aws-authz-smart/pull/63)! I'm going to go ahead and close this issue for now, but please feel free to reopen or reach out if the issue is not resolved.

PR #63, opened 2021-09-09
fix: remove filtered responses from total count

ssvegaraju:
Issue #, if available:
#55
Description of changes:
Updated the `authorizeAndFilterReadResponse` method to correctly update the total number of results in a bundle when filtering unauthorized resources.

### DIFF
--- a/src/smartHandler.test.ts
+++ b/src/smartHandler.test.ts
@@ -734,6 +734,7 @@ describe('authorizeAndFilterReadResponse', () => {
             lastUpdated: '2020-11-20T11:10:48.034+00:00',
         },
         type: 'searchset',
+        total: 0,
         link: [
             {
                 relation: 'self',
@@ -749,6 +750,7 @@ describe('authorizeAndFilterReadResponse', () => {
     const searchAllEntitiesMatch = {
         ...emptySearchResult,
         entry: [createEntry(validPatient), createEntry(validPatientObservation), createEntry(validPatientEncounter)],
+        total: 3,
     };
 
     const searchSomeEntitiesMatch = {
@@ -760,6 +762,7 @@ describe('authorizeAndFilterReadResponse', () => {
             createEntry({ ...validPatientObservation, subject: 'not-you' }),
             createEntry(validPatientEncounter),
         ],
+        total: 5,
     };
     const searchNoEntitiesMatch = {
         ...emptySearchResult,
@@ -767,6 +770,7 @@ describe('authorizeAndFilterReadResponse', () => {
             createEntry({ ...validPatient, id: 'not-yours' }),
             createEntry({ ...validPatientObservation, subject: 'not-you' }),
         ],
+        total: 2,
     };
     const cases: (string | ReadResponseAuthorizedRequest | boolean | any)[][] = [
         [
@@ -998,7 +1002,7 @@ describe('authorizeAndFilterReadResponse', () => {
                 readResponse: searchAllEntitiesMatch,
             },
             true,
-            { ...emptySearchResult, entry: [createEntry(validPatientEncounter)] },
+            { ...emptySearchResult, entry: [createEntry(validPatientEncounter)], total: 1 },
         ],
         [
             'SEARCH: user scope; Practitioner able to search and get ALL results',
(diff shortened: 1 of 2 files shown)


src/smartHandler.ts
```diff
@@ -329,6 +329,11 @@ export class SMARTHandler implements Authorization {
                     this.fhirVersion,
                 ),
             );
+            const totalNumEntriesAfterFilter: number =
+                readResponse.total - (readResponse.entry.length - entries.length);
+            if (totalNumEntriesAfterFilter < readResponse.total) {
+                readResponse.total = totalNumEntriesAfterFilter;
+            }
```
rsmayda: Do you need this `if` block?

src/smartHandler.ts
```diff
@@ -329,6 +329,11 @@ export class SMARTHandler implements Authorization {
                     this.fhirVersion,
                 ),
             );
+            const totalNumEntriesAfterFilter: number =
```
rsmayda: Can you add tests for this?

src/smartHandler.ts
ssvegaraju: Some of the tests were failing to due `readResponse.total` being NaN, but I changed the if condition to be clearer on that.

src/smartHandler.ts
ssvegaraju: I added a few tests for this but realized that they were redundant as we already have tests filtering results from the searches both partially and completely.

rsmayda: It looks like the existing tests are not taking into account the `total`: https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/mainline/src/smartHandler.test.ts#L730

Could you change the existing tests to look more of what the router sends in (i.e. include the `total`):
https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/bundle/bundleGenerator.ts#L25

src/smartHandler.ts
```diff
@@ -329,6 +329,9 @@ export class SMARTHandler implements Authorization {
                     this.fhirVersion,
                 ),
             );
+            if (!Number.isNaN(readResponse.total + 1)) {
+                readResponse.total -= readResponse.entry.length - entries.length;
+            }
```
rsmayda: The router should always send the total: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/bundle/bundleGenerator.ts#L32

but if it doesn't lets set it to the existing length:

```suggestion
            if (!readResponse.total) {
                readResponse.total = entries.length;
            } else {
                readResponse.total -= readResponse.entry.length - entries.length;
            }
```

rsmayda (changes requested):

src/smartHandler.ts
ssvegaraju: Updated all of the tests for this module to include totals as an adjusted parameter similar to how `entry` was processed!

src/smartHandler.ts
ssvegaraju: Implemented this change, but did it a bit differently as directly modifying `readResponse.total` proved to be a bad idea due to the fact that the test runs `authorizeAndFilterReadResponse` twice per filter (one with the FhirServiceBaseUrl and one without), and the request would end up modified from the previous call. Instead, passing it back as part of the spread operator worked.

src/smartHandler.test.ts
```diff
@@ -760,13 +762,15 @@ describe('authorizeAndFilterReadResponse', () => {
             createEntry({ ...validPatientObservation, subject: 'not-you' }),
             createEntry(validPatientEncounter),
         ],
+        total: 5,
```
rsmayda, line 765: `searchSomeEntitiesMatch` should have a differing output total looks like total should be set to 3? I don't see that being tested - maybe I missed it though?

this is the case name: `SEARCH: user scope; Patient's history results are filtered to only contain valid entries`

rsmayda (changes requested):

src/smartHandler.test.ts
ssvegaraju: Yeah, the total is actually 3 because it uses `searchAllEntitiesMatch` as the `respBody` in the test cases (the 4th parameter in that case definition on line 974). This has a total of 3.

rsmayda: Thank you for explaining!

rsmayda (approved):

PR #64, opened 2021-09-09
fix: scope checking for getting status on an export

rsmayda:
Description of changes:
- Change scope checking on non - `initiate-export` operations
  - It will now allow if system or user scope
  - This does not change the requirement that the job owner MUST match the requestor id
- removed `['*', 'read'].includes(smartScope.accessType)` check as it was duplicative

### DIFF
--- a/src/smartScopeHelper.ts
+++ b/src/smartScopeHelper.ts
@@ -84,30 +84,37 @@ function isSmartScopeSufficientForBulkDataAccess(
     smartScope: ClinicalSmartScope,
     scopeRule: ScopeRule,
 ) {
-    let bulkDataRequestHasCorrectScope = false;
-    if (bulkDataAuth.exportType === 'system') {
-        bulkDataRequestHasCorrectScope =
-            ['system', 'user'].includes(smartScope.scopeType) &&
-            smartScope.resourceType === '*' &&
-            ['*', 'read'].includes(smartScope.accessType) &&
-            getValidOperationsForScopeTypeAndAccessType(
-                smartScope.scopeType,
-                smartScope.accessType,
-                scopeRule,
-            ).includes('read');
-    } else if (bulkDataAuth.exportType === 'group') {
-        bulkDataRequestHasCorrectScope =
-            ['system'].includes(smartScope.scopeType) &&
-            ['*', 'read'].includes(smartScope.accessType) &&
-            getValidOperationsForScopeTypeAndAccessType(
-                smartScope.scopeType,
-                smartScope.accessType,
-                scopeRule,
-            ).includes('read');
+    if (bulkDataAuth.operation === 'initiate-export') {
+        let bulkDataRequestHasCorrectScope = false;
+        if (bulkDataAuth.exportType === 'system') {
+            bulkDataRequestHasCorrectScope =
+                ['system', 'user'].includes(smartScope.scopeType) &&
+                smartScope.resourceType === '*' &&
+                ['*', 'read'].includes(smartScope.accessType) &&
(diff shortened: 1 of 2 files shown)


src/smartScopeHelper.ts
```diff
@@ -84,30 +84,24 @@ function isSmartScopeSufficientForBulkDataAccess(
...
+                ['system', 'user'].includes(scopeType) && resourceType === '*' && hasReadPermissions;
+        } else if (bulkDataAuth.exportType === 'group') {
+            bulkDataRequestHasCorrectScope = ['system'].includes(scopeType) && hasReadPermissions;
+        }
+        return bulkDataRequestHasCorrectScope;
     }
     return (
-        ['initiate-export', 'get-status-export', 'cancel-export'].includes(bulkDataAuth.operation) &&
-        bulkDataRequestHasCorrectScope
+        ['get-status-export', 'cancel-export'].includes(bulkDataAuth.operation) &&
```
Bingjiling, line 102: get-status-export will return the pre-signed URL for the exported data if the export is completed. Would this result in a requester not having '*' resource access to get the system export data?

src/smartScopeHelper.ts
```diff
@@ -84,30 +84,24 @@ function isSmartScopeSufficientForBulkDataAccess(
...
+    const { scopeType, accessType, resourceType } = smartScope;
+    const hasReadPermissions = getValidOperationsForScopeTypeAndAccessType(scopeType, accessType, scopeRule).includes(
+        'read',
+    );
+    if (bulkDataAuth.operation === 'initiate-export') {
+        let bulkDataRequestHasCorrectScope = false;
+        if (bulkDataAuth.exportType === 'system') {
+            bulkDataRequestHasCorrectScope =
+                ['system', 'user'].includes(scopeType) && resourceType === '*' && hasReadPermissions;
+        } else if (bulkDataAuth.exportType === 'group') {
```
Bingjiling, line 96: Maybe we can map the get status request to the correct exportType through the DDB field exportType in the export table? This way we don't need special handling for 'get-status-export' and 'cancel-export' and requests are mapped to the correct type.

rsmayda: This is unlikely as this is called in the `verifyAccessToken` method which is the first AuthZ/N method called (prior to getting any info from DDB).

The another approach is to have a sperate `get-status` url for `group` exports

src/smartScopeHelper.ts
rsmayda: Yes, that is correct. IMO the resource type check `*` is not effective and does not translate to the `group` export either. The export scopes matter when you initiate the export. 

After initiation you are just getting an export

src/smartScopeHelper.ts
Bingjiling: Gotcha. Yea this approach makes sense as long as it does not create security issues.

src/smartScopeHelper.ts
rsmayda: As an aside the FHIR spec does not specify the scope behavior for `status` or `delete`

https://hl7.org/fhir/uv/bulkdata/export/index.html#bulk-data-status-request

Bingjiling: Gotcha! I just checked we do verify the requester getting job status is the same as the owner of the job [here](https://github.com/awslabs/fhir-works-on-aws-routing/blob/666a4c472426d8b0aad5651c70aae8f605d7dc84/src/router/handlers/exportHandler.ts#L38) so we are be good here. Thanks for verifying it in the FHIR spec.

Bingjiling (approved):
LGTM

Issue #78, opened 2022-02-24
[Bug] _include issue when system scopes are not set to *


ghost:
**Describe the bug**
An issue occurs when we have specific system scopes specified (Ex. system/DocumentReference.read and system/Patient.read, no system/*.read it will only return the resource that was requested and it filters out all of the included resources.
 
Example: We request a DocumentReference and do an _include to get the Patient resource.
 
If we have system/DocumentReference.read and system/Patient.read as scopes, it will only return the DocumentReference resource. If we have system/*.read it will return both. It appears to be in the getValidOperationsForScope function in smartScopeHelper.ts.  The hasAccessToResource function filters the included Patient out because only DocumentReference is in the usable scopes.

**Expected behavior**
The expected result would be for it to check against the full list of scopes to see if it was valid.

ssvegaraju:
Hi @medhost-gkersting, just wanted to reach out to let you know that a fix for this issue has been merged in! Thanks for bringing up this issue, I'll go ahead close it but feel free to reopen if there's anything else!

PR #79, opened 2022-03-03
fix: scopes should not be filtered in search use cases

ssvegaraju:
Issue #78 , if available:

Description of changes:
Updated the function checking valid operations for scope to process search use cases similar to transactions/batches. This means that searches running include will contain all expected resources and not drop scopes to limit search results since they are not directly requested.

**Testing**
To test that the functionality was working as intended, I deployed on SMART and created a `DocumentReference` with a reference to a `Patient` resource, and searched with the example scopes (`system/DocumentReference.read`, `system/Patient.read`) to reproduce the issue. I then implemented the fix as well as an additional unit test, and made sure that the Patient resource was included in the search results.

### DIFF
--- a/src/smartScopeHelper.test.ts
+++ b/src/smartScopeHelper.test.ts
@@ -405,7 +405,7 @@ describe('filterOutUnusableScope', () => {
                 clonedScopeRule,
                 'search-type',
                 false,
-                'DocumentReference'
+                'DocumentReference',
             ),
         ).toEqual(['system/DocumentReference.read', 'system/Patient.read']);
     });
--- a/src/smartScopeHelper.test.ts
+++ b/src/smartScopeHelper.test.ts
@@ -395,6 +395,19 @@ describe('filterOutUnusableScope', () => {
             ),
         ).toEqual(['system/*.write']);
     });
+    
+    test('filter patient in search use case', () => {
+        const clonedScopeRule = emptyScopeRule();
+        clonedScopeRule.system.read = ['search-type'];
+        expect(
+            filterOutUnusableScope(
+                ['system/DocumentReference.read', 'system/Patient.read'],
+                clonedScopeRule,
+                'search-type',
+                false,
+            ),
+        ).toEqual(['system/DocumentReference.read', 'system/Patient.read']);
+    });
 });
 
 describe('getValidOperationsForScopeTypeAndAccessType', () => {
--- a/src/smartScopeHelper.test.ts
+++ b/src/smartScopeHelper.test.ts
@@ -395,7 +395,7 @@ describe('filterOutUnusableScope', () => {
             ),
         ).toEqual(['system/*.write']);
     });
-    
+
     test('filter patient in search use case', () => {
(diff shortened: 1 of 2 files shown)


src/smartScopeHelper.test.ts
```diff
@@ -395,6 +395,19 @@ describe('filterOutUnusableScope', () => {
             ),
         ).toEqual(['system/*.write']);
     });
+
+    test('filter patient in search use case', () => {
+        const clonedScopeRule = emptyScopeRule();
+        clonedScopeRule.system.read = ['search-type'];
+        expect(
+            filterOutUnusableScope(
```
Bingjiling, line 403: In the situation mentioned in the issue, `reqResourceType` should be `Patient` right?

ssvegaraju: I might have misunderstood, but I think the situation was that the Patient resource was a reference in the DocumentReference, and needed to be included in the results with the `_include` parameter. This test is making sure that neither scope is filtered out from a search, so that resources of both `DocumentReference` and `Patient` can be read by the search query. Additionally, if the `reqResourceType` parameter is provided, the route is different and won't actually hit the change made to fix the bug: https://github.com/awslabs/fhir-works-on-aws-authz-smart/pull/79/files#diff-9539c459709b0437ecf87b68dce629addd51e10c3c4f7af6ab3fdfcc171a312aL57

rsmayda: Yeah for the test case we need to ensure: `reqResourceType?: string`, is set

src/smartScopeHelper.test.ts
```diff
@@ -395,6 +395,19 @@ describe('filterOutUnusableScope', () => {
             ),
         ).toEqual(['system/*.write']);
     });
+
+    test('filter patient in search use case', () => {
```
rsmayda:
```suggestion
    test('do not filter patient scope out in type-search use case', () => {
```

src/smartScopeHelper.test.ts
ssvegaraju: Yup, after running tests searches do indeed provide the `reqResourceType` parameter, whereas batch and transaction do not. So I moved the change into the correct place to reflect that. I also updated the test to set `reqResourceType` and it correctly returns both scopes as valid.

rsmayda (approved):
Great research!

liquid36:
Hi! I'm having an issue with this change.  
This is my use case: 

```javascript
        const clonedScopeRule = emptyScopeRule();
        clonedScopeRule.system.read = ['search-type'];
        expect(
            filterOutUnusableScope(
                ['system/DocumentReference.read', 'system/Patient.read'],
                clonedScopeRule,
                'search-type',
                false,
                'Procedure',
            ),
        ).toEqual([]);
```
If the user claims for this scope `system/DocumentReference.read system/Patient.read` but ask for a **Procedure** resource, the `filterOutUnusableScope` mark both scope as usable and then the operation is allowed.

Issue #83, opened 2022-04-06
Scopes are not filtered on search-type


liquid36:
Hi! I'm having an issue in this lines which changes on this   #79 

https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/93df88ff3fe07adaf65a28f0244259f719c7066b/src/smartScopeHelper.ts#L58

This is my use case: 

```javascript
        const clonedScopeRule = emptyScopeRule();
        clonedScopeRule.system.read = ['search-type'];
        expect(
            filterOutUnusableScope(
                ['system/DocumentReference.read', 'system/Patient.read'],
                clonedScopeRule,
                'search-type',
                false,
                'Procedure',
            ),
        ).toEqual([]);
```
If the user claims for this scope `system/DocumentReference.read system/Patient.read` but ask for a **Procedure** resource, the `filterOutUnusableScope` mark both scope as usable and then the operation is allowed.

ssvegaraju:
Hey @liquid36, thanks for catching this! I've added it to our team's backlog.

ghost:
We are also experiencing this bug with user/ scopes.

It allows a user access to ANY resource type as long as there is at least one user/ resource in the scope list.

Example:  A scope list with user/Procedure.read
  "scp": [
    "openid",
    "fhirUser",
    "offline_access",
    "user/Procedure.read"
  ],

allows access to read Observations

FernandoAranda:
Hi @medhost-bfindeisen , thank you for bringing this to our attention, 
we are working on this issue and is already in our backlog.

-Fernando

rsmayda:
@medhost-bfindeisen is this specifically for `search-type`? or for `read` as well

ghost:
@rsmayda , this appears specific to search-type, read would result in the last condition here to be false:

https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/mainline/src/smartScopeHelper.ts#L58

rsmayda:
@medhost-bfindeisen we weren't able to replicate with the `user` scope could you send over the jwt `scp` claims, the `fhirUser` claim and your request? I wonder if it is related to this input param: https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/mainline/src/smartHandler.ts#L74

ghost:
Hi @rsmayda 

Our scp only has access to Procedure.read:
```
  "scp": [
    "openid",
    "fhirUser",
    "offline_access",
[...]

PR #88, opened 2022-09-08
fix: only allow scopes to be returned iff it matches resourceType

rsmayda:
Issue #, if available: #83

Description of changes: Checks to ensure that there is at least 1 valid scope against the resourceType for `search-type` requests. If there isn't the jwt validation will fail

### DIFF
--- a/src/smartHandler.test.ts
+++ b/src/smartHandler.test.ts
@@ -429,6 +429,16 @@ describe('verifyAccessToken', () => {
             { ...baseAccessNoScopes, scp: 'system/Patient.write' },
             true,
         ],
+        [
+            'patientUserSystem_specificRead_search',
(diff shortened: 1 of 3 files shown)


ssvegaraju (approved):
I think this looks good! Just to confirm I understood correctly:
If a search performed with (rev)/include (e.g. we request a DocumentReference and do an _include to get the Patient resource, then as long as we have the scope for the DocumentReference, the patient scopes won't be filtered out, correct?

rsmayda:
> I think this looks good! Just to confirm I understood correctly: If a search performed with (rev)/include (e.g. we request a DocumentReference and do an _include to get the Patient resource, then as long as we have the scope for the DocumentReference, the patient scopes won't be filtered out, correct?

That's right!

PR #92, opened 2022-09-12
fix: handle include/revInclude correctly

rsmayda:
Issue #, if available: #83

Description of changes:
- Moving the logic of dealing with non-root resourceTypes returned from a `search-type` operation from `filterOutUnusableScope` to `authorizeAndFilterReadResponse`
- Make `hasSystemAccess` more explicit
- Improve test coverage

### DIFF
--- a/src/smartAuthorizationHelper.test.ts
+++ b/src/smartAuthorizationHelper.test.ts
@@ -542,7 +542,7 @@ describe('introspectJwtToken', () => {
             Math.floor(Date.now() / 1000),
             Math.floor(Date.now() / 1000) + 10,
             expectedAudValue,
-            `${expectedIssValue}/`,
+            `${expectedIssValue}`,
         );
         const mock = new MockAdapter(axios);
         mock.onPost(introspectUrl).reply(200, {
--- a/src/smartAuthorizationHelper.ts
+++ b/src/smartAuthorizationHelper.ts
@@ -118,9 +118,9 @@ export function isFhirUserAdmin(fhirUser: FhirResource, adminAccessTypes: string
 }
 
 /**
- * @param scopes this should be scope set from the `verifyAccessToken` method
- * @param resourceType the type of the resource the request is trying to access
- * @param accessModifier the type of access the request is asking for
+ * @param scopes: this should be scope set from the `verifyAccessToken` method
+ * @param resourceType: the type of the resource the request is trying to access
+ * @param accessModifier: the type of access the request is asking for
  * @returns if there is a usable system scope for this request
  */
 export function hasSystemAccess(scopes: string[], resourceType: string, accessModifier: AccessModifier): boolean {
@@ -135,6 +135,7 @@ export function hasSystemAccess(scopes: string[], resourceType: string, accessMo
             );
         } catch (e) {
             // Error occurs from `convertScopeToSmartScope` if scope was invalid
+            logger.debug((e as any).message);
             return false;
         }
     });
--- a/src/smartAuthorizationHelper.ts
+++ b/src/smartAuthorizationHelper.ts
@@ -8,7 +8,9 @@ import { decode, verify } from 'jsonwebtoken';
 import axios from 'axios';
 import resourceReferencesMatrixV4 from './schema/fhirResourceReferencesMatrix.v4.0.1.json';
 import resourceReferencesMatrixV3 from './schema/fhirResourceReferencesMatrix.v3.0.1.json';
-import { FhirResource, IntrospectionOptions } from './smartConfig';
+import { AccessModifier, FhirResource, IntrospectionOptions } from './smartConfig';
+import { convertScopeToSmartScope } from './smartScopeHelper';
+
 import getComponentLogger from './loggerBuilder';
 
 export const FHIR_USER_REGEX =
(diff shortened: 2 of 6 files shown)


src/smartHandler.test.ts
```diff
@@ -1212,20 +1219,79 @@ describe('authorizeAndFilterReadResponse', () => {
...
+                    scopes: ['system/Observation.read', 'system/Encounter.read'],
+                    usableScopes: ['system/Observation.read'],
+                },
+                operation: 'search-type',
+                readResponse: searchAllEntitiesMatch,
+            },
+            true,
+            searchFilteredEntitiesMatch,
+        ],
+        [
```
Bingjiling, line 1279: NitPick: The test would be more readable if each case is represented as an object with informative field names v.s. an array.

rsmayda: Good feedback - I will keep this in mind. Since this is the paradigm currently adopted I will leave as is but created a clean up ticket

src/smartAuthorizationHelper.test.ts
```diff
@@ -542,7 +542,7 @@ describe('introspectJwtToken', () => {
             Math.floor(Date.now() / 1000),
             Math.floor(Date.now() / 1000) + 10,
             expectedAudValue,
-            `${expectedIssValue}/`,
+            `${expectedIssValue}`,
```
ssvegaraju, line 545: Is there a reason we removed the `/` or is it just a cleanup thing?

kharbutli-meta:
Other than the unit tests, were there any other tests run to verify functionality?

src/smartHandler.ts
```diff
@@ -378,6 +392,7 @@ export class SMARTHandler implements Authorization {
                 this.adminAccessTypes,
                 fhirServiceBaseUrl,
                 this.fhirVersion,
+                'write',
```
ssvegaraju, line 396: Do we not want to do the same validation as for read requests above here for write requests? I know the bug was mainly introduced with the search feature, but is there any chance it could be replicated for write through a Bundle or something?

src/smartScopeHelper.test.ts
```diff
@@ -118,24 +118,36 @@ describe.each(isScopeSufficientCases)('ScopeType: %s: isScopeSufficient', (scope
...
+    });
     test('scope is insufficient to do a transaction', () => {
         const clonedScopeRule = emptyScopeRule();
         clonedScopeRule[scopeType].read = ['read'];
         clonedScopeRule[scopeType].write = ['create'];
 
         expect(isScopeSufficient(`${scopeType}/*.*`, clonedScopeRule, 'transaction', false)).toEqual(false);
     });
-    test('invalid scope', () => {
+    test('invalid; `read` scope with no resourcType', () => {
```
ssvegaraju: nit: spelling `resourceType`

src/smartScopeHelper.test.ts
```diff
@@ -433,10 +464,12 @@ describe('filterOutUnusableScope', () => {
                 'search-type',
                 false,
                 'Practitioner',
+                undefined,
+                'patient',
             ),
         ).toEqual([]);
     });
-    test('do not filter patient scope out in type-search use case', () => {
+    test('SYSTEM Scope: do filter Patient resourceType scope out in type-search use case', () => {
```
ssvegaraju, line 472: Does this reversion mean revinclude/include works a different way now? (i.e. are the changes now that we check in the authorizeAndFilterReadResponse to see if they have access to a search resource rather than with the initial request?)

src/smartAuthorizationHelper.ts
```diff
@@ -116,14 +118,26 @@ export function isFhirUserAdmin(fhirUser: FhirResource, adminAccessTypes: string
 }
 
 /**
- * @param usableScopes this should be usableScope set from the `verifyAccessToken` method
- * @param resourceType the type of the resource we are trying to access
+ * @param scopes this should be scope set from the `verifyAccessToken` method
```
kharbutli-meta: nit: Add a ':' after the name of the param to make the description more readable.

src/smartAuthorizationHelper.ts
```diff
@@ -116,14 +118,26 @@ export function isFhirUserAdmin(fhirUser: FhirResource, adminAccessTypes: string
...
+ * @param resourceType the type of the resource the request is trying to access
```
kharbutli-meta: ditto

src/smartAuthorizationHelper.ts
```diff
@@ -116,14 +118,26 @@ export function isFhirUserAdmin(fhirUser: FhirResource, adminAccessTypes: string
...
+ * @param accessModifier the type of access the request is asking for
```
kharbutli-meta: ditto

src/smartAuthorizationHelper.ts
```diff
@@ -116,14 +118,26 @@ export function isFhirUserAdmin(fhirUser: FhirResource, adminAccessTypes: string
...
+        try {
+            const clinicalSmartScope = convertScopeToSmartScope(scope);
+
+            return (
+                clinicalSmartScope.scopeType === 'system' &&
+                (clinicalSmartScope.resourceType === '*' || clinicalSmartScope.resourceType === resourceType) &&
+                (clinicalSmartScope.accessType === '*' || clinicalSmartScope.accessType === accessModifier)
+            );
+        } catch (e) {
+            // Error occurs from `convertScopeToSmartScope` if scope was invalid
```
kharbutli-meta, line 137: Should an error be logged? An error is not always benevolent.

src/smartScopeHelper.ts
```diff
@@ -172,16 +172,5 @@ export function filterOutUnusableScope(
             ),
     );
 
-    // We should only return the scopes iff there is at least 1 valid scope for the given resourceType
```
kharbutli-meta, line 175: Why are we removing this logic instead of updating it?

src/smartAuthorizationHelper.test.ts
rsmayda: clean up this was a broken test

src/smartScopeHelper.test.ts
rsmayda: no it will have the same behavior;; `usableScope` will now just be for the request i.e. if 
- scopes are `patient/Patient.read patient/Observation.read` 
- the request is `/Patient?name=Bob&_include=Observation`

this will result in passing `verifyToken` method and `usablescope` only being `patient/Patient.read`. Then in `authorizeAndFilterReadResponse` it will use the remaining scopes to validate the Observation resources 

if request was`/Medication?_include=Observation` then it would fail `verifyToken` since there would be no `usableScope`

ssvegaraju: Gotcha, that makes sense!

src/smartScopeHelper.ts
rsmayda: With the removal of `search-type` from this line https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/80b690760d7c85bda996bdb5f0c273acb1a20005/src/smartScopeHelper.ts#L61

We will have the same behavior described in this comment

src/smartHandler.ts
rsmayda: Added more test to address this comment, let me know your thoughts

ssvegaraju: I think it looks good!

rsmayda:
> Other than the unit tests, were there any other tests run to verify functionality?

I was able to deploy and test locally

ssvegaraju (approved):
Lgtm

kharbutli-meta (approved):

PR #103, opened 2022-11-22
test: combo tests for getSearchFilterBasedOnIdentity

kcadette:
Issue #, if available:

Description of changes:


package.json
```diff
@@ -30,7 +30,10 @@
   },
   "dependencies": {
     "axios": "^0.21.4",
+    "csv-writer": "^1.6.0",
```
Bingjiling: These dependencies are already in dev dependencies. Please remove them from this clause.

ssvegaraju: These should be in devDependencies, right?

src/comboTest/getAllowedResourceTypesForOperation-combo.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        };
+        testCases.push([JSON.stringify(testCase, null, 2), testCase]);
+    });
+    return testCases;
+};
+
+describe('getAllowedResourceTypesForOperation-combo', () => {
+    const testResults: any[] = [];
+    const keysToOutput: any[] = [
+        { field: 'testName', title: 'Test Number' },
```
Bingjiling: Based on what we talked about this afternoon, patient in context and fhir user is needed for reviewing the result as well. e.g. if the only scope is patient/*.read but there's no patient in context, then this function should return empty instead of all resource types.

src/comboTest/getSearchFilterBasedOnIdentity-combo.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+    csv.forEach((inputRow, index) => {
+        const testCase: any = {};
+
+        const row = inputRow.csvRow;
+        testCase.testName = `Combo Test Row ${index}`;
+        testCase.request = {
+            userIdentity: inputRow.userIdentity,
+            fhirServiceBaseUrl: testStubs.convertToBaseUrl(row.fhirServiceBaseUrl),
+            // operation: row.operation,
```
ssvegaraju: should operation be commented out here?

src/comboTest/getSearchFilterBasedOnIdentity-combo.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        };
+        testCase.rawCsvRow = row;
+        testCases.push([JSON.stringify(testCase, null, 2), testCase]);
+    });
+    return testCases;
+};
+
+describe('getSearchFilterBasedOnIdentity-combo', () => {
+    const testResults: any[] = [];
+    const keysToOutput: any[] = [
```
ssvegaraju, line 59: could we order this so that the Usable Scopes and Scopes arrays are last?

ssvegaraju (commented):
Overall looks good! Few minor comments, and looks like we have another `.DS_Store` file that got committed.

src/comboTest/getSearchFilterBasedOnIdentity-combo.test.ts
Bingjiling: Remove it if it's not needed?

src/comboTest/getSearchFilterBasedOnIdentity-combo.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        { field: 'testName', title: 'Test Number' },
+        { field: 'rawCsvRow.fhirUser', title: 'FHIR User' },
+        { field: 'rawCsvRow.patientContext', title: 'Patient Context' },
+        { field: 'rawCsvRow.fhirServiceBaseUrl', title: 'Base Url' },
+        { field: 'request.resourceType', title: 'Resource Type' },
+        { field: 'request.userIdentity.usableScopes', title: 'Usable Scopes' },
+        { field: 'request.userIdentity.scopes', title: 'Scopes' },
```
Bingjiling: Let's use the scopes from the rawCsvRow, even though they should be exactly the same in this case, in principle we should be reviewing the original input before any processing. The same goes for the other test file.

src/comboTest/getSearchFilterBasedOnIdentity-combo.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        '4.0.1',
+        undefined,
+        undefined,
+        true,
+    );
+
+    test.each(testCases)('CASE: %s', async (testCaseString, testCase) => {
+        let testResult: any;
+        try {
+            // console.log(testCase.request.userIdentity.fhirUserObject);
```
Bingjiling: Remove if not needed.

src/comboTest/getSearchFilterBasedOnIdentity-combo.test.ts
```diff
@@ -0,0 +1,101 @@
...
+            testResult = await authZHandler.getSearchFilterBasedOnIdentity(
+                <GetSearchFilterBasedOnIdentityRequest>testCase.request,
+            );
+            console.log(testResult);
```
Bingjiling: Remove for hygiene.

Bingjiling (commented):
It's a good practice to create PR focusing on only one thing -- it enables faster review/response and also keeps each commit laser focused for tracking and later audit if needed. In this case specifically, it'll be good to have two PRs each having one test instead of in one PR. Let's get this one going as is, but keep this in mind for future PRs where each PR address one thing.

Bingjiling:
Also please remove the .DS_Store file and update the .gitignore file to keep it from happening agin if needed. (I can't find a good place to leave comment for it in the Files Changed section. )

src/comboTest/getAllowedResourceTypesForOperation-combo.test.ts
kcadette: This isn't true
This function is only using scopes passed in to determine AllowedResourceTypes. Isn't this why I am validating that this function is always used with another authorizing function?
The only time a wildcard would not return all resource types is when the request operation is not a part of the scopeRule.
But we did discuss patient in context and fhir user for the getSearchFilter functions. They are both included there

kcadette:
I originally had 2 PRs but consolidated them. Ill reopen the other one

src/comboTest/getAllowedResourceTypesForOperation-combo.test.ts
Bingjiling: Yes you're right! I definitely mixed those two up.

Bingjiling (approved):
LGTM

PR #107, opened 2022-12-07
feat: reject mixed system scope

Bingjiling:
Issue #, if available:

Description of changes:
* Reject system scope mixed with user or patient scope 
* Update unit tests
* Remove dependency from combo test to smartHandlerTest file to avoid duplicative tests run

### DIFF
--- a/src/smartHandler.ts
+++ b/src/smartHandler.ts
@@ -24,6 +24,9 @@ import { JwksClient } from 'jwks-rsa';
 import { SMARTConfig, UserIdentity } from './smartConfig';
 import {
     convertScopeToSmartScope,
+    FHIR_PATIENT_SCOPE_REGEX,
+    FHIR_SYSTEM_SCOPE_REGEX,
+    FHIR_USER_SCOPE_REGEX,
     filterOutUnusableScope,
     getScopes,
     getValidOperationsForScopeTypeAndAccessType,
@@ -153,7 +156,7 @@ export class SMARTHandler implements Authorization {
             }
             if (
                 !usableScopes.some((scope: string) => {
-                    return scope.startsWith('system');
+                    return FHIR_SYSTEM_SCOPE_REGEX.test(scope);
                 })
             ) {
                 // if requestor is relying on the "user" scope we need to verify they are coming from the correct endpoint & resourceType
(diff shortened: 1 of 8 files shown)


src/smartScopeHelper.test.ts
```diff
@@ -561,3 +562,19 @@ describe('convertScopeToSmartScope', () => {
...
+        expect(() => {
+            rejectInvalidScopeCombination(['system/Patient.read', 'user/Organization.read']);
+        }).toThrowError(UnauthorizedError);
+    });
+    test('reject system scope mixed with patient scope', () => {
+        expect(() => {
+            rejectInvalidScopeCombination(['system/Patient.read', 'patient/Observation.read']);
+        }).toThrowError(UnauthorizedError);
+    });
+    test('allow patient scope mixed with user scope', () => {
```
rsmayda, line 577: Could you add a test for a single system scope and a multiple system scopes

src/smartScopeHelper.ts
```diff
@@ -145,6 +145,17 @@ export function isScopeSufficient(
     return false;
 }
 
+export function rejectInvalidScopeCombination(scopes: string[]): void {
+    if (
+        scopes.some((scope: string) => scope.startsWith('system')) &&
+        (scopes.some((scope: string) => scope.startsWith('user')) ||
+            scopes.some((scope: string) => scope.startsWith('patient')))
```
rsmayda:
```suggestion
        scopes.some((scope: string) => scope.startsWith('system/')) &&
        (scopes.some((scope: string) => scope.startsWith('user/')) ||
            scopes.some((scope: string) => scope.startsWith('patient/')))
```

src/smartScopeHelper.ts
```diff
@@ -145,6 +145,17 @@ export function isScopeSufficient(
...
+    ) {
+        console.log(scopes);
```
rsmayda: Are we using a logger in FWoA?
`const logger = getComponentLogger();`

src/smartHandler.ts
```diff
@@ -124,6 +125,7 @@ export class SMARTHandler implements Authorization {
 
         // get just the scopes that apply to this request
         const scopes = getScopes(decodedToken[this.config.scopeKey]);
+        rejectInvalidScopeCombination(scopes);
         const usableScopes = filterOutUnusableScope(
```
rsmayda, line 129: on this line but on line 156 could you change `return scope.startsWith('system');` to `return scope.startsWith('system/');`

src/smartScopeHelper.ts
rsmayda: Better yet lets not do this but make use of Regex (safer option), can we make this change across all 'startsWith' usage?

Bingjiling: Sure, I can make that change.

src/smartScopeHelper.ts
Bingjiling: Oops, this was added for debugging. Will remove.

rsmayda (approved):
Please get second review

ssvegaraju (approved):
LGTM!

